PR #4, opened 2023-05-23
HALSim Component: Device

gabryelreyes:
Added Device and SocketClient


lib/ConvoyLeader/src/App.cpp
```diff
@@ -60,10 +61,19 @@
 
 void App::setup()
 {
+    if (false == Board::getInstance().init())
```
BlueAndi, line 64: It would be better if Initialization of the HAL doesn't fail. I see that its because the SocketClient init() will do the connection establishment. But wouldn't it be better to have the connection establishment with possible reconnect behaviour during processing?

lib/HALSim/src/Board.h
```diff
@@ -0,0 +1,233 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+/*******************************************************************************
+    DESCRIPTION
+*******************************************************************************/
+/**
+ * @brief  The simulation board realization.
+ * @author Andreas Merkle <web@blue-andi.de>
```
BlueAndi, line 29: You are the author.

lib/HALSim/src/Board.h
```diff
@@ -0,0 +1,233 @@
...
+/******************************************************************************
+ * Macros
+ *****************************************************************************/
+
+/******************************************************************************
+ * Types and Classes
+ *****************************************************************************/
+
+/**
+ * The concrete simulation robot board.
```
BlueAndi, line 62: robot?

lib/HALSim/src/Board.h
```diff
@@ -0,0 +1,233 @@
...
+     *
+     * @returns If all components are correctly initialized, returns true. Otherwise, false.
+     */
+    bool init() final
+    {
+        bool isReady = false;
+
+        if (false == m_device.init())
+        {
+            /* Handle Device error */
```
BlueAndi, line 90: Missing implementation.

lib/HALSim/src/Board.h
```diff
@@ -0,0 +1,233 @@
...
+     *
+     * @returns If all components are correct, returns true. Otherwise, false;
+     */
+    bool process() final
+    {
+        bool isSuccess = false;
+
+        if (false == m_device.process())
+        {
+            /* Handle Device error */
```
BlueAndi, line 112: Missing implementation.

lib/HALSim/src/Board.h
```diff
@@ -0,0 +1,233 @@
...
+    /** Button driver */
+    Button m_button;
+
+    /** Device driver */
+    Device m_device;
+
+    /** Red LED driver */
+    LedBlue m_ledBlue;
+
+    /** Red LED driver */
```
BlueAndi, line 207: Green

lib/HALSim/src/Board.h
```diff
@@ -0,0 +1,233 @@
...
+    /** Battery driver */
+    Battery m_battery;
+
+    /** Button driver */
+    Button m_button;
+
+    /** Device driver */
+    Device m_device;
+
+    /** Red LED driver */
```
BlueAndi, line 204: Blue

lib/HALSim/src/Button.cpp
```diff
@@ -0,0 +1,92 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+/*******************************************************************************
+    DESCRIPTION
+*******************************************************************************/
+/**
+ * @brief  Button realization
+ * @author Andreas Merkle <web@blue-andi.de>
```
BlueAndi, line 29: You are the author ... please this finding is relevant for all other files too.

lib/HALSim/src/Button.h
```diff
@@ -0,0 +1,101 @@
...
+
+/******************************************************************************
+ * Macros
+ *****************************************************************************/
+
+/******************************************************************************
+ * Types and Classes
+ *****************************************************************************/
+
+/** This class provides access to the robot simulation button. */
```
BlueAndi, line 56: robot?

lib/HALSim/src/Button.h
```diff
@@ -0,0 +1,101 @@
...
+     * Constructs the button adapter.
+     */
+    Button() : IButton()
+    {
+    }
+
+    /**
+     * Destroys the button adapter.
+     */
+    ~Button()
```
BlueAndi, line 70: virtual is missing

lib/HALSim/src/Device.cpp
```diff
@@ -0,0 +1,101 @@
...
+/* Default server port of the device. */
+const char* Device::DEFAULT_SERVER_PORT = "65432";
+
+/******************************************************************************
+ * Public Methods
+ *****************************************************************************/
+
+bool Device::init()
+{
+    return m_socket.init(DEFAULT_SERVER_ADDRESS, DEFAULT_SERVER_PORT);
```
BlueAndi, line 69: Connection establishment in a initialization routine is dangerous. What about doing this in the process() routine?

lib/HALSim/src/Device.h
```diff
@@ -0,0 +1,119 @@
...
+
+/******************************************************************************
+ * Macros
+ *****************************************************************************/
+
+/******************************************************************************
+ * Types and Classes
+ *****************************************************************************/
+
+/** This class provides access to the robot simulation device. */
```
BlueAndi, line 57: robot?

lib/HALSim/src/Device.h
```diff
@@ -0,0 +1,119 @@
...
+     * Constructs the device adapter.
+     */
+    Device() : IDevice(), m_socket()
+    {
+    }
+
+    /**
+     * Destroys the device adapter.
+     */
+    ~Device()
```
BlueAndi, line 71: virtual is missing.

lib/HALSim/src/LedGreen.h
```diff
@@ -0,0 +1,89 @@
...
+     * Constructs the green LED adapter.
+     */
+    LedGreen() : ILed()
+    {
+    }
+
+    /**
+     * Destroys the green LED adapter.
+     */
+    ~LedGreen()
```
BlueAndi, line 70: virtual is missing

BlueAndi (changes requested):

lib/HALSim/src/Battery.cpp
```diff
@@ -0,0 +1,87 @@
...
+ *****************************************************************************/
+
+/******************************************************************************
+ * Public Methods
+ *****************************************************************************/
+
+uint32_t Battery::getVoltage()
+{
+    /* Not implemented. */
+    return 0U;
```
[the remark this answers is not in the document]
gabryelreyes: A 0 would be considered undervoltage and therefore the application should go into error state. It was set to 0 until the error state, as well as the voltage parameters are defined, i.e., undervoltage, nominal voltage, etc. In any case, a 0 is considered unvalid/unsafe and the application shall stop.
Same principle applies for getChargeLevel()

lib/ConvoyLeader/src/App.cpp
gabryelreyes: The connection of the SocketClient has been moved to the process rutine. However part of the initialization can still fail, for example, the context creation. Other components can fail, for example, the battery. For this reason, the HAL should fail in the setup.

This makes also sense in the Target environment, where the USB Host Chip can fail to initialize, or WiFi can fail to connect.

lib/ConvoyLeader/src/App.cpp
```diff
@@ -60,10 +61,19 @@
...
+        while (true)
+        {
+            ;
+        }
+    }
 }
 
 void App::loop()
 {
+    (void)Board::getInstance().process();
```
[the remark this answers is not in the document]
gabryelreyes: Battery, Device and Network can all fail. The behaviour is going into an error state. If the error comes all the way back to the application, is considered fatal.

Logging module is not implemented in this branch, but these errors will also be logged once it gets merged.

lib/HALSim/src/Board.h
gabryelreyes: Errors will be logged once the Logging Module gets merged

PR #6, opened 2023-06-02
Add Network Component

gabryelreyes:


lib/HALSim/src/Board.h
```diff
@@ -114,6 +121,12 @@ class Board : public IBoard
             /* Log Device error */
             LOG_ERROR("Device failed to connect. ");
```
BlueAndi, line 122: You don't know here whether a connection attempt failed or something other.

BlueAndi (changes requested):

BlueAndi (approved):

PR #12, opened 2023-09-22
HAL for RemoteControl App

gabryelreyes:
HAL changes for RemoteControl App


lib/ArduinoNative/src/Arduino.cpp
```diff
@@ -129,3 +169,93 @@ extern void delay(unsigned long ms)
 /******************************************************************************
  * Local Functions
  *****************************************************************************/
+
+#ifdef UNIT_TEST
+
+/**
+ * Handle the Arguments passed to the programm.
+ *
+ * @param[in] argc Program argument count
```
BlueAndi, line 178: prgArguments is missing

lib/ArduinoNative/src/Arduino.cpp
```diff
@@ -129,3 +169,93 @@ extern void delay(unsigned long ms)
 /******************************************************************************
  * Local Functions
  *****************************************************************************/
+
+#ifdef UNIT_TEST
+
+/**
+ * Handle the Arguments passed to the programm.
```
BlueAndi, line 176: lowercase --> arguments

lib/ArduinoNative/src/WString.h
```diff
@@ -466,6 +466,44 @@ class String
         return isEmptyFlag;
     }
 
+    /**
+     * Remove all characters from the String tarting at the index until the end of the String.
```
BlueAndi, line 470: typo: starting

lib/ArduinoNative/src/WString.h
```diff
@@ -466,6 +466,44 @@ class String
         return isEmptyFlag;
     }
 
+    /**
+     * Remove all characters from the String tarting at the index until the end of the String.
+     * @param[in] index Index at which to remove characters.
+     */
+    void remote(unsigned int index)
```
BlueAndi, line 473: typo: remove not remote :-)

lib/ArduinoNative/src/WString.h
```diff
@@ -466,6 +466,44 @@ class String
...
+        {
+            return;
+        }
+        if (count > length() - index)
+        {
+            count = length() - index;
+        }
+        unsigned int newlen = length() - count;
+        char*        tmp    = new char[newlen + 1];
+        memcpy(tmp, m_buffer, index);
```
BlueAndi, line 499: Did you get the buffer?

lib/ArduinoNative/src/WiFi.h
```diff
@@ -0,0 +1,81 @@
...
+     * @returns MAC Address.
+     */
+    String macAddress()
+    {
+        return "";
+    }
+};
+
+/** Extern WiFi Class. */
+WiFi_ WiFi;
```
BlueAndi, line 75: Oh ... this means everywhere you include this header, it will create the instance. Might not be good.

lib/HALInterfaces/src/INetwork.h
```diff
@@ -89,15 +89,21 @@ class INetwork
...
+     * @param[in] birthTopic    Birth topic. If empty, no birth message is used.
+     * @param[in] birthMessage  Birth message.
      * @param[in] willTopic     Last will topic. If empty, no last will is used.
      * @param[in] willMessage   Last will message.
      * @param[in] reconnect     If true, the client will try to reconnect to the broker, if the connection is lost.
      * @return If successfully set, returns true. Otherwise, false.
      */
-    virtual bool setConfig(const String& clientId, const String& brokerAddress, uint16_t brokerPort,
-                           const String& willTopic, const String& willMessage, bool reconnect) = 0;
+    virtual bool setConfig(const String& clientId, const String& ssid, const String& password,
```
BlueAndi, line 103: We should think about using a struct.

lib/HALSim/src/Device.h
```diff
@@ -115,6 +130,12 @@ class Device : public IDevice
      * Socket stream for communication with the device.
      */
     SocketClient m_socket;
+
+    /** Server Address. */
+    const char* m_address;
+
+    /** Server Port Number. */
+    const char* m_port;
```
BlueAndi, line 138: Why do we use a string instead of uint16_t?

lib/HALSim/src/FileReader.cpp
```diff
@@ -0,0 +1,91 @@
...
+    {
+        size_t readBytes = fread(outBuffer, sizeof(char), maxBufferSize, file);
+
+        if (ferror(file) != 0)
+        {
+            LOG_ERROR("Error ocurred while reading configuration file.");
+        }
+        else
+        {
+            outBuffer[readBytes++] = '\0'; /* Just to be safe. */
```
BlueAndi, line 76: bad style having the post-operator inside the braces.
What happens if you read maxBufferSize?

lib/HALSim/src/FileReader.cpp
```diff
@@ -0,0 +1,91 @@
...
+    bool  isSuccess = false;
+    FILE* file      = fopen(fileName.c_str(), "r");
+
+    if (nullptr == file)
+    {
+        LOG_ERROR("Error ocurred while opening configuration file \"%s\".", fileName.c_str());
+    }
+    else
+    {
+        size_t readBytes = fread(outBuffer, sizeof(char), maxBufferSize, file);
```
BlueAndi, line 68: What happens if the file is bigger than the maxBufferSize?

lib/HALSim/src/Network.cpp
```diff
@@ -155,6 +158,12 @@ bool Network::setConfig(const String& clientId, const String& brokerAddress, uin
     }
     else
     {
+        if (false == birthTopic.isEmpty())
```
BlueAndi, line 161: Is that correct or shouldn't we just assign it? Fact is, you can never clear it after there is something in the m_birthTopic.

lib/HALSim/src/Network.cpp
```diff
@@ -155,6 +158,12 @@ bool Network::setConfig(const String& clientId, const String& brokerAddress, uin
     }
     else
     {
+        if (false == birthTopic.isEmpty())
+        {
+            m_birthTopic   = birthTopic;
+            m_birthMessage = birthMessage;
+        }
+
         if (false == willTopic.isEmpty())
```
BlueAndi, line 167: Same question as above.

lib/HALSim/src/Network.cpp
```diff
@@ -203,8 +212,17 @@ bool Network::connect()
     {
         LOG_DEBUG("MQTT client connected to broker");
         resubscribe();
-        m_state   = STATE_CONNECTED;
-        isSuccess = true;
+        m_state = STATE_CONNECTED;
```
BlueAndi, line 215: Is the mosquitto_connect() blocking until the connection is established?

lib/HALSim/src/SocketClient.cpp
```diff
@@ -344,6 +344,13 @@ bool SocketClient::process()
                         {
                             m_members->m_rcvQueue.push(static_cast<uint8_t>(recvbuf[idx]));
                         }
+
+                        if (CANARY_WARNING_LIMIT < m_members->m_rcvQueue.size())
+                        {
+                            printf("Warning: SocketClient receive queue is getting full. %d bytes "
```
BlueAndi, line 350: print() ?

lib/HALSim/src/SocketClient.h
```diff
@@ -201,6 +201,9 @@ class SocketClient : public Stream
     /** SocketClient Members. PIMPL Idiom. */
     SocketClientImpl* m_members;
 
+    /** Maximum amount of available Bytes allowed before issuing a warning. */
+    static const uint16_t CANARY_WARNING_LIMIT = 500U;
```
BlueAndi, line 205: Pls move the constant before the member variables.

lib/HALTarget/src/FileReader.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+/******************************************************************************
+ * Public Methods
+ *****************************************************************************/
+
+bool FileReader::readFile(const String& fileName, char* outBuffer, const uint32_t maxBufferSize)
+{
+    bool isSuccess = false;
+
+    if (false == LittleFS.begin(true))
```
BlueAndi, line 62: The filesystem shall be mounted once.

lib/HALTarget/src/FileReader.cpp
```diff
@@ -0,0 +1,102 @@
...
+            LOG_ERROR("Failed to open file \"%s\".", fileName.c_str());
+        }
+        else
+        {
+            uint32_t bytesRead = 0U;
+            bytesRead          = file.readBytes(outBuffer, maxBufferSize);
+
+            if (0U == bytesRead)
+            {
+                LOG_ERROR("Error ocurred while reading configuration file.");
```
BlueAndi, line 81: How does the FileReader knows what I would like to load?

lib/HALTarget/src/FileReader.cpp
```diff
@@ -0,0 +1,102 @@
...
+        if ((false == file) || (file.isDirectory()))
+        {
+            LOG_ERROR("Failed to open file \"%s\".", fileName.c_str());
+        }
+        else
+        {
+            uint32_t bytesRead = 0U;
+            bytesRead          = file.readBytes(outBuffer, maxBufferSize);
+
+            if (0U == bytesRead)
```
BlueAndi, line 79: 0 is valid.

lib/HALTarget/src/Network.cpp
```diff
@@ -82,14 +89,15 @@ Network::~Network()
...
     m_state = STATE_IDLE;
-
     return true;
 }
 
 bool Network::process()
 {
     bool isSuccess = true;
 
+    isSuccess = manageWiFi();
```
BlueAndi, line 99: Assign in the line above. There is no need to assign "true" first.

lib/HALTarget/src/Network.cpp
```diff
@@ -134,13 +148,21 @@ bool Network::setConfig(const String& clientId, const String& brokerAddress, uin
     }
     else
     {
+        if (false == birthTopic.isEmpty())
```
BlueAndi, line 151: Same questions than in the Sim HAL.

lib/HALTarget/src/Network.cpp
```diff
@@ -196,6 +218,16 @@ bool Network::connect()
         LOG_DEBUG("MQTT client connected to broker");
```
BlueAndi, line 218: Isnt't that asynchronous?

lib/HALTarget/src/Network.cpp
```diff
@@ -316,13 +368,16 @@ void Network::unsubscribe(const String& topic)
...
 {
-    if ((false == m_clientId.isEmpty()) && (false == m_brokerAddress.isEmpty()) && (0U != m_brokerPort))
+    if ((false == m_clientId.isEmpty()) && (false == m_brokerAddress.isEmpty()) && (0U != m_brokerPort) &&
+        (false == m_wiFiSSID.isEmpty()))
     {
         (void)m_mqttClient.setServer(m_brokerAddress.c_str(), m_brokerPort);
         (void)m_mqttClient.setBufferSize(MAX_BUFFER_SIZE);
         (void)m_mqttClient.setKeepAlive(MQTT_KEEP_ALIVE_S);
         (void)m_mqttClient.setCallback([this](char* topic, uint8_t* payload, uint32_t length)
                                        { this->onMessageCallback(topic, payload, length); });
```
BlueAndi, line 378: Does all these functions have return values?

BlueAndi (changes requested):

lib/HALSim/src/Device.h
gabryelreyes: Because it is given as string from the command line, and winsock also needs it as string. Unnecessary conversion to and from uint

lib/ArduinoNative/src/WString.h
gabryelreyes: Sorry, I do not understand the question

lib/ArduinoNative/src/WiFi.h
gabryelreyes: Would "static" be a good solution in this case?

lib/HALInterfaces/src/INetwork.h
gabryelreyes: You are right. I would make this a new issue and resolve it later.

lib/HALSim/src/Network.cpp
gabryelreyes: It is not blocking but it is also not async. The function returns an error if not able to connect, without waiting. There is also an async method but requires refactoring of the loop.

lib/HALSim/src/Network.cpp
gabryelreyes: With this there is the possibility of setting a custom birth/will topic from the application, or none at all. But if this is not necessary, then we can directly assign it.

lib/HALSim/src/SocketClient.cpp
gabryelreyes: Sorry I do not understand the question

lib/HALTarget/src/FileReader.cpp
gabryelreyes: But an empty file is useless (?)

lib/HALSim/src/FileReader.cpp
gabryelreyes: Is it a good idea to add this null byte? Or is it better to remove this line altogether?

lib/HALSim/src/FileReader.cpp
gabryelreyes: Solved by checking EOF

lib/HALTarget/src/Network.cpp
gabryelreyes: I do not think so: https://pubsubclient.knolleary.net/api#connect

lib/HALTarget/src/Network.cpp
gabryelreyes: Some return the client instance, allowing the function to be chained

lib/ArduinoNative/src/WString.h
BlueAndi: NULL pointer check is missing, you can not be sure that you got the buffer.

lib/HALTarget/src/Network.cpp
BlueAndi: The API doesn't describe whether its asynchronous or not. Thats because it depends on the client which is used!

lib/HALSim/src/SocketClient.cpp
BlueAndi: Why printf instead of LOG_...?

lib/HALTarget/src/FileReader.h
```diff
@@ -0,0 +1,95 @@
...
+class FileReader : public IFileReader
+{
+public:
+    /**
+     * Get FileReader instance.
+     * @returns FileReader instance.
+     */
+    static FileReader& getInstance()
+    {
+        static FileReader instance; /* Idiom. */
```
BlueAndi, line 62: Why shall this be a singleton?

BlueAndi (changes requested):

BlueAndi (changes requested):

lib/HALTarget/src/FileReader.h
gabryelreyes: It does not nedd to be a singleton

BlueAndi (approved):

PR #25, opened 2023-11-05
Snippets

gabryelreyes:
Added code Snippets and Templates, as well as format-on-save setting.


.vscode/templates.code-snippets
```diff
@@ -0,0 +1,204 @@
...
+		"description": "Simple Doxygen comment."
+	},
+	"Doxygen Block Comment/Description": {
+		"prefix": "doxy_block",
+		"scope": "c, cpp",
+		"body": [
+			"/**",
+			" * ${1:description}",
+			" * ",
+			" * ${2:parameters}",
```
BlueAndi, line 50: usually one empty line between parameters and return description

.vscode/templates.code-snippets
```diff
@@ -0,0 +1,204 @@
...
+		],
+		"description": "Doxygen comment/description block."
+	},
+	"Template Header File": {
+		"prefix": "template_header",
+		"scope": "c, cpp",
+		"body": [
+			"/* MIT License",
+			" *",
+			" * Copyright (c) 2023 Andreas Merkle <web@blue-andi.de>",
```
BlueAndi, line 63: Can the year be automatically set to the current year.

BlueAndi (changes requested):

.vscode/templates.code-snippets
gabryelreyes: Not automatically, no. We could make it variable, and each time we have to type the year. Or we change the template each year. Or make an script of some kind to run on the CI

gabryelreyes: Forget it, just found out how

BlueAndi (approved):

PR #31, opened 2023-11-07
Remote otadev

chezlyne:
Added WebServerCustom, Upload and FileManager.


BlueAndi:
@chezlyne The CI failed, please fix the code correspondingly.

BlueAndi:
@chezlyne There are two similar branches. Which one is valid?
image

lib/ZumoOta/src/WebServerCustom.h
```diff
@@ -0,0 +1,55 @@
...
+ * @{
+ */
+
+#ifndef WebServerCustom_H
+#define WebServerCustom_H
+
+/******************************************************************************
+ * Includes
+ *****************************************************************************/
+#include <Arduino.h>
```
nhjschulz, line 19: include in header only what is needed by it. All others need to go into cpp

lib/ZumoOta/src/FileManager.cpp
```diff
@@ -0,0 +1,205 @@
...
+FileManager::FileManager() {
+    
+}
+    
+
+FileManager::~FileManager() {
+}
+   
+
+bool FileManager::createUploadPage() {
```
nhjschulz, line 70: The file is not to be created by code on the fly, Then you don't need to put it into the FS. You have it as string already.

Figure out how to upload a filesystem in similar way as hex file.  Code must handle a missing file system as an error during startup.

lib/ZumoOta/src/App.cpp
```diff
@@ -34,7 +34,10 @@
  *****************************************************************************/
 #include "App.h"
 #include <Arduino.h>
+#include <WiFi.h>
 #include <ESPAsyncWebServer.h>
```
nhjschulz, line 38: why this, should be imported through your custom webserver include already.

lib/ZumoOta/src/FileManager.cpp
```diff
@@ -0,0 +1,205 @@
...
+        Serial.println("Fehler beim Öffnen der Upload-Seite.");
+        return "";
+    }
+    
+    String pageContent = uploadPage.readString();
+    uploadPage.close();
+
+    return pageContent;
+
+    /*if (!LittleFS.begin()) {
```
nhjschulz, line 178: commented code - why ?  Remove or comment why it is there.

lib/ZumoOta/src/App.cpp
```diff
@@ -83,12 +95,22 @@ void App::setup()
...
 
-    server.on("/", HTTP_GET, [](AsyncWebServerRequest *request){
-        request->send(200, "text/plain", "Hello, world");
-    });
-
-    server.begin();
+    webserver.begin();
+    fileManager.createUploadPage();
+    
+    String pageContent = fileManager.readUploadPage();
```
nhjschulz, line 101: move into line 106

lib/ZumoOta/src/App.h
```diff
@@ -59,14 +61,14 @@ class App
 {
 public:
     /**
-     * Construct the convoy leader application.
+     * Construct the app  application.
      */
     App()
     {
     }
```
nhjschulz, line 68: suprised that it works with the {} here. Should be ony App(); as implemented in cpp

lib/ZumoOta/src/FileManager.h
```diff
@@ -0,0 +1,70 @@
...
+ */
+#ifndef FILEMANAGER_H
+#define FILEMANAGER_H
+
+/******************************************************************************
+ * Includes
+ *****************************************************************************/
+#include <Arduino.h>
+#include <FS.h>
+#include <LittleFS.h>
```
nhjschulz, line 20: LittleFS is implementation detail. Move inlcude to cpp

lib/ZumoOta/src/App.cpp
```diff
@@ -83,12 +95,22 @@ void App::setup()
...
+    
+    if (!pageContent.isEmpty()) { 
+      
+        webserver.on("/", HTTP_GET, [pageContent](AsyncWebServerRequest *request) {
```
nhjschulz, line 105: the webserver config should move into your custom webserver class, or otherwise it has no real value. From app it should only be webserver.init() or similar.

chezlyne: I will perform it with the next commit.

lib/ZumoOta/src/FileManager.cpp
chezlyne: FileManager is only to send the webpages to the client. What you describe here will be tacle with further developments in the Uploader.

nhjschulz: The entire FileManager can be deleted based on discussion in sync meeting from 09.11.2023. Webserver can directly deliver content from file system. No need for own code to load it.

chezlyne: I will use it just to initialize the FileSystem.

lib/ZumoOta/src/App.h
```diff
@@ -82,9 +81,15 @@ class App
...
-/******************************************************************************
- * Functions
- *****************************************************************************/
+    /**
+     * Start the Application.
+     */
+    void start();
+
+private:
+    FileManager fileManager{};
```
nhjschulz, line 90: Remove the {}    from the variables

nhjschulz, line 90: Prefix member variables with "m_". I saw that Gabryel follows this convention, so it should be consistent.

nhjschulz, line 90: Variables need commenting

lib/ZumoOta/src/FileManager.cpp
```diff
@@ -0,0 +1,85 @@
...
+FileManager::FileManager()
+{
+}
+    
+
+FileManager::~FileManager()
+{
+}
+   
+void FileManager::init()
```
nhjschulz, line 71: void is not a good return type if init can fail. use bool and return true for success, false otherwise.

lib/ZumoOta/src/FileManager.h
```diff
@@ -0,0 +1,48 @@
...
+ * @author Decareme Pauline Ngangnou <ngandeca@yahoo.fr>
+ * @{
+ */
+#ifndef FILEMANAGER_H
+#define FILEMANAGER_H
+
+/******************************************************************************
+ * Includes
+ *****************************************************************************/
+#include <Arduino.h>
```
nhjschulz, line 15: why do you include this here ? Remove and add to cpp if needed there.

lib/ZumoOta/src/Upload.h
```diff
@@ -0,0 +1,55 @@
...
+ * @{
+ */
+
+#ifndef UPLOAD_H
+#define UPLOAD_H
+
+/******************************************************************************
+ * Includes
+ *****************************************************************************/
+#include <stdint.h>
```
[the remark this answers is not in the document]
nhjschulz: comment above still not resolved

lib/ZumoOta/src/WebServerCustom.h
nhjschulz: comment above still open

lib/ZumoOta/src/App.cpp
```diff
@@ -83,17 +101,13 @@ void App::setup()
...
     Serial.println(WiFi.localIP());
 
-    server.on("/", HTTP_GET, [](AsyncWebServerRequest *request){
-        request->send(200, "text/plain", "Hello, world");
-    });
-
-    server.begin();
-
+    fileManager.init();
+    start();
```
nhjschulz, line 105: is this start really needed in addition to setup ? 
What is your rule to do things in setup() or in start() ? At least for me it is unclear why filemanager init() is here, but webServer.init() in start()

nhjschulz (changes requested):

lib/ZumoOta/src/Upload.h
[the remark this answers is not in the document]
chezlyne: I will perform it with the next commit.

lib/ZumoOta/src/WebServerCustom.h
chezlyne: I will perform it with the next commit.

lib/ZumoOta/src/App.h
chezlyne: I will perform it with the next commit.

chezlyne: I will perform it with the next commit.

chezlyne: I will perform it with the next commit.

lib/ZumoOta/src/FileManager.cpp
chezlyne: I will perform it with the next commit.

lib/ZumoOta/src/FileManager.h
chezlyne: I will perform it with the next commit.

lib/ZumoOta/src/App.cpp
chezlyne: I will modify the start() Method by the next commit. Start() Method will be responsible for initializing and starting the application. It will primarily involves initializing the FileManager and WebServerCustom components.

nhjschulz (approved):

PR #36, opened 2023-11-21
Remote otadev

chezlyne:

PR #45, opened 2023-11-27
Sync RemoteOTABasic with main branch

gabryelreyes:
Not sure if this should be done before or after #36 
That PR could benefit from the current state of Settings and Network components


nhjschulz:
I'm not seeing a conflict with #36, but would give #36 priority as this one is huge and you are more experienced to solve potential merge issues. At lesst sync with owner of #36 before merging.

gabryelreyes:
We will wait until #36 is done and then I will merge and fix any problems in this PR

PR #51, opened 2023-11-30
Update RemoteOTA

gabryelreyes:
- Removed custom implementations in favor of HAL components.
- Setup and Loop refactored to use HAL components and SettingsHandler, as well as to be similar to other applications.
- Some files have been visually re-formatted during the update.
- Added Doxygen documentation generation.
- Findings in the Documentation shall be handled by the author of the ZumoOta application @chezlyne . This may be done before or after this PR has been merged. The reviewers shall decide when this is to be done.


gabryelreyes:
Formatted all relevant files in 8f958ea22f49faac35410fe22074b2f6893e0b4e
This may make it difficult to review the changes of this PR, but will make it easier in future PRs as automatic formatting has been activated.

nhjschulz (approved):
thanks for the tidy-up.

We should think about a Base class for App that handles WebSever, Board ... 
Otherwise this dublicates across variants.

PR #58, opened 2023-12-05
Preparation for V2VClient

gabryelreyes:
Preparation for the introduction of the V2VClient

### DIFF
--- a/lib/ConvoyLeader/src/App.cpp
+++ b/lib/ConvoyLeader/src/App.cpp
@@ -217,10 +217,10 @@ void App::setup()
                         processingChainFactory.registerLateralSafetyPolicyCreateFunc(LateralSafetyPolicy::create);
 
                         if (false == m_platoonController.init(
-                                         [this](Waypoint& waypoint) { inputWaypointCallback(waypoint); },
-                                         [this](const Waypoint& waypoint) { outputWaypointCallback(waypoint); },
+                                         [this](Waypoint& waypoint) { return inputWaypointCallback(waypoint); },
+                                         [this](const Waypoint& waypoint) { return outputWaypointCallback(waypoint); },
                                          [this](const int16_t left, const int16_t right)
-                                         { motorSetpointCallback(left, right); }))
+                                         { return motorSetpointCallback(left, right); }))
                         {
                             LOG_FATAL("Could not initialize Platoon Controller.");
                         }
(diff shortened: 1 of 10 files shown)


lib/ConvoyLeader/src/App.cpp
```diff
@@ -216,10 +204,10 @@ void App::setup()
                         processingChainFactory.registerLateralSafetyPolicyCreateFunc(LateralSafetyPolicy::create);
 
                         if (false == m_platoonController.init(
-                                         [this](Waypoint& waypoint) { inputWaypointCallback(waypoint); },
-                                         [this](const Waypoint& waypoint) { outputWaypointCallback(waypoint); },
+                                         [this](Waypoint& waypoint) { return inputWaypointCallback(waypoint); },
```
BlueAndi, line 207: Move the lambdas to variables and assign the variables here. Otherwise it looks quite strange.

lib/HALInterfaces/src/IMqttClient.h
```diff
@@ -139,18 +139,20 @@ class IMqttClient
...
      * Subscribes to a topic.
      *
-     * @param[in] topic     Topic to subscribe to. The Client ID is used as base topic: `Client ID`/`topic`
+     * @param[in] topic     Topic to subscribe to.
+     * @param[in] useClientBaseTopic   If true, the client ID is used as the base (prefix) of the topic.
      * @param[in] callback  Callback function, which is called on a new message.
      * @return If successfully subscribed, returns true. Otherwise, false.
      */
-    virtual bool subscribe(const String& topic, TopicCallback callback) = 0;
+    virtual bool subscribe(const String& topic, const bool useClientBaseTopic, TopicCallback callback) = 0;
```
BlueAndi, line 147: Now its no generic MQTT client anymore. Shouldn't that be done on higher level?

gabryelreyes: I believe it is the other way around. Before, the client ID was used always, which makes it not-generic MQTT Client. It could ONLY subscribe to topics of the form `clientID/topic`

Now it is optional to include the client ID as base topic, so it can subscribe to any topic.

It is left as an option to use the client ID for publishing and subscribing, as the application may not know the MQTT client ID

gabryelreyes:
Force push required to fix conflict with line-endings. Sorry about that.

PR #59, opened 2023-12-05
V2V Client

gabryelreyes:
- Implemented V2V client
- IDs loaded from config.json file
- IDs set per Command Line arguments

### DIFF
--- a/data/config/config.json
+++ b/data/config/config.json
@@ -1,23 +1,23 @@
 {
     "robotName": "",
-    "WIFI": {
-        "SSID": "TheForce",
-        "PSWD": "hanshotfirst"
+    "wifi": {
+        "ssid": "TheForce",
+        "pswd": "hanshotfirst"
     },
-    "MQTT": {
-        "HOST": "192.168.137.1",
-        "PORT": 1883
+    "mqtt": {
+        "host": "192.168.137.1",
+        "port": 1883
     },
-    "AP": {
-        "SSID": "DCS_AP",
-        "PSWD": "hanshotfirst"
+    "ap": {
+        "ssid": "DCS_AP",
+        "pswd": "hanshotfirst"
     },
-    "WEBSERVER": {
-        "USER": "admin",
-        "PSWD": "admin"
+    "webServer": {
+        "user": "admin",
+        "pswd": "admin"
     },
-    "PLATOON": {
-        "PLATOON_ID": 0,
-        "VEHICLE_ID": 0
+    "platoon": {
+        "platoonId": 0,
+        "vehicleId": 0
     }
 }
--- a/data/config/config.json
+++ b/data/config/config.json
@@ -15,5 +15,9 @@
     "WEBSERVER": {
         "USER": "admin",
         "PSWD": "admin"
+    },
+    "PLATOON": {
+        "PLATOON_ID": 0,
+        "VEHICLE_ID": 0
     }
 }
--- a/lib/ArduinoNative/src/Arduino.cpp
+++ b/lib/ArduinoNative/src/Arduino.cpp
@@ -131,7 +131,7 @@ static const char PRG_ARG_RADON_ULZER_ADDR_DEFAULT[] = "localhost";
 static const char PRG_ARG_RADON_ULZER_PORT_DEFAULT[] = "65432";
 
 /** Program argument default value of the configuration file. */
-static const char PRG_ARG_CFG_FILE_DEFAULT[] = "data/config/config.json";
+static const char PRG_ARG_CFG_FILE_DEFAULT[] = "./data/config/config.json";
 
 /** Program argument default value of the Platoon ID. */
 static const char PRG_ARG_PLATOON_ID_DEFAULT[] = "0";
--- a/lib/ArduinoNative/src/Arduino.cpp
+++ b/lib/ArduinoNative/src/Arduino.cpp
@@ -318,10 +318,13 @@ extern void delay(unsigned long ms)
     }
 }
 
+#ifndef UNIT_TEST
+
 extern bool convertToJson(const String& str, JsonVariant variant)
 {
     return variant.set(str.c_str());
 }
+#endif
 
 /******************************************************************************
  * Local Functions
--- a/lib/ArduinoNative/src/Arduino.cpp
+++ b/lib/ArduinoNative/src/Arduino.cpp
@@ -41,6 +41,7 @@
 #include <Device.h>
 #include <getopt.h>
 #include <direct.h>
+#include <ConfigurationKeys.h>
 #endif
 
 /******************************************************************************
@@ -158,52 +159,6 @@ static const char WEBSERVER_USER_DEFAULT[] = "admin";
 
 /** Default value of the webserver passphrase. */
 static const char WEBSERVER_PASSPHRASE_DEFAULT[] = "admin";
-
-/**
- * Configuration keys found in the config file.
- */
-namespace configKeys
-{
-    /** Robot name */
-    static const char ROBOT_NAME[] = "robotName";
-
-    /** WiFi */
-    static const char WIFI[] = "wifi";
-
-    /** MQTT */
-    static const char MQTT[] = "mqtt";
-
-    /** Access point */
-    static const char AP[] = "ap";
-
-    /** WebServer */
-    static const char WEBSERVER[] = "webServer";
-
-    /** Platoon */
-    static const char PLATOON[] = "platoon";
-
-    /** SSID */
-    static const char SSID[] = "ssid";
-
-    /** Passphrase */
-    static const char PASSWORD[] = "pswd";
-
-    /** Host */
-    static const char HOST[] = "host";
-
-    /** Port */
-    static const char PORT[] = "port";
-
-    /** User */
-    static const char USER[] = "user";
-
-    /** Platoon ID */
-    static const char PLATOON_ID[] = "platoonId";
-
-    /** Vehicle ID */
-    static const char VEHICLE_ID[] = "vehicleId";
-
-} // namespace configKeys
 
 #endif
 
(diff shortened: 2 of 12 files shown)


lib/ConvoyLeader/src/App.cpp
```diff
@@ -180,6 +180,15 @@ void App::setup()
                 {
                     LOG_FATAL("MQTT configuration could not be set.");
                 }
+                else if (0U != settings.getPlatoonVehicleId())
```
BlueAndi, line 183: Please use a static const variable like PLATTOON_LEADER_ID instead of just 0 here.

lib/PlatoonService/src/V2VClient.cpp
```diff
@@ -0,0 +1,267 @@
...
+static const uint32_t JSON_BIRTHMESSAGE_MAX_SIZE = 64U;
+
+/******************************************************************************
+ * Public Methods
+ *****************************************************************************/
+
+V2VClient::V2VClient(MqttClient& mqttClient) :
+    m_mqttClient(mqttClient),
+    m_waypointQueue(),
+    m_inputTopic(""),
```
BlueAndi, line 79: Remove "" from strings.

lib/PlatoonService/src/V2VClient.cpp
```diff
@@ -0,0 +1,267 @@
...
+V2VClient::~V2VClient()
+{
+}
+
+bool V2VClient::init(uint8_t platoonId, uint8_t vehicleId)
+{
+    bool    isSuccessful = false;
+    char    inputTopicBuffer[MAX_TOPIC_LENGTH];
+    char    outputTopicBuffer[MAX_TOPIC_LENGTH];
+    uint8_t outputTopicId = vehicleId + 1U;
```
BlueAndi, line 94: Why +1?

lib/PlatoonService/src/V2VClient.cpp
```diff
@@ -0,0 +1,267 @@
...
+
+    if (0U == vehicleId)
```
BlueAndi, line 96: Magic number

lib/PlatoonService/src/V2VClient.cpp
```diff
@@ -0,0 +1,267 @@
...
+        LOG_DEBUG("Input Topic: %s", m_inputTopic.c_str());
+        LOG_DEBUG("Output Topic: %s", m_outputTopic.c_str());
+
+        if ((true == m_inputTopic.isEmpty()) || (true == m_outputTopic.isEmpty()))
+        {
+            LOG_FATAL("Failed to create Platoon MQTT topics.");
+        }
+        /* Subscribe to Input Topic. */
+        else if (false == m_mqttClient.subscribe(m_inputTopic, false,
+                                                 [this](const String& payload)
```
BlueAndi, line 141: Please use variables for the lambdas.

lib/PlatoonService/src/V2VClient.cpp
```diff
@@ -0,0 +1,267 @@
...
+        /* Set topics. */
+        m_inputTopic  = inputTopicBuffer;
+        m_outputTopic = outputTopicBuffer;
+
+        LOG_DEBUG("Input Topic: %s", m_inputTopic.c_str());
+        LOG_DEBUG("Output Topic: %s", m_outputTopic.c_str());
+
+        if ((true == m_inputTopic.isEmpty()) || (true == m_outputTopic.isEmpty()))
+        {
+            LOG_FATAL("Failed to create Platoon MQTT topics.");
```
BlueAndi, line 137: Just to remember ... fatal means dead .... no way out anymore.

lib/PlatoonService/src/V2VClient.cpp
```diff
@@ -0,0 +1,267 @@
...
+    jsonPayload["Y"]           = waypoint.yPos;        /**< Y position [mm]. */
+    jsonPayload["Orientation"] = waypoint.orientation; /**< Orientation [mrad]. */
+    jsonPayload["Left"]        = waypoint.left;        /**< Left motor speed [steps/s]. */
+    jsonPayload["Right"]       = waypoint.right;       /**< Right motor speed [steps/s]. */
+    jsonPayload["Center"]      = waypoint.center;      /**< Center speed [steps/s]. */
+
+    size_t jsonBufferSize = measureJson(jsonPayload) + 1U;
+    char   jsonBuffer[jsonBufferSize];
+
+    if (0U == serializeJson(jsonPayload, jsonBuffer, jsonBufferSize))
```
BlueAndi, line 176: Should we compare it against jsonBufferSize?

lib/Service/src/SettingsHandler.cpp
```diff
@@ -152,15 +164,17 @@ bool SettingsHandler::saveConfigurationFile(const String& filename)
...
-    doc["WEBSERVER"]["PSWD"] = m_webServerPassword.c_str();
+    doc["robotName"]             = m_robotName.c_str();
+    doc["WIFI"]["SSID"]          = m_wifiSSID.c_str();
+    doc["WIFI"]["PSWD"]          = m_wifiPassword.c_str();
+    doc["MQTT"]["HOST"]          = m_mqttBrokerAddress.c_str();
+    doc["MQTT"]["PORT"]          = m_mqttPort;
+    doc["AP"]["SSID"]            = m_apSSID.c_str();
+    doc["AP"]["PSWD"]            = m_apPassword.c_str();
+    doc["WEBSERVER"]["USER"]     = m_webServerUser.c_str();
+    doc["WEBSERVER"]["PSWD"]     = m_webServerPassword.c_str();
```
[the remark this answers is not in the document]
gabryelreyes: But it requires methods not present in the ArduinoNative implementation of String

data/config/config.json
```diff
@@ -15,5 +15,9 @@
     "WEBSERVER": {
         "USER": "admin",
         "PSWD": "admin"
+    },
+    "PLATOON": {
+        "PLATOON_ID": 0,
```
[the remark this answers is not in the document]
gabryelreyes: `platoonId ` and `vehicleId` instead of all uppercase then? Should WEBSERVER be changed too?

lib/ArduinoNative/src/Arduino.cpp
```diff
@@ -446,11 +483,17 @@ static int createConfigFile(const PrgArguments& prgArgs)
...
+                jsonDoc["WIFI"]["SSID"]          = WIFI_SSID_DEFAULT;
+                jsonDoc["WIFI"]["PSWD"]          = WIFI_PASSPHRASE_DEFAULT;
+                jsonDoc["MQTT"]["HOST"]          = prgArgs.mqttHost;
+                jsonDoc["MQTT"]["PORT"]          = prgArgs.mqttPort;
+                jsonDoc["AP"]["SSID"]            = AP_SSID_DEFAULT;
+                jsonDoc["AP"]["PSWD"]            = AP_PASSPHRASE_DEFAULT;
+                jsonDoc["WEBSERVER"]["USER"]     = WEBSERVER_USER_DEFAULT;
+                jsonDoc["WEBSERVER"]["PSWD"]     = WEBSERVER_PASSPHRASE_DEFAULT;
+                jsonDoc["PLATOON"]["PLATOON_ID"] = prgArgs.platoonId;
+                jsonDoc["PLATOON"]["VEHICLE_ID"] = prgArgs.vehicleId;
```
nhjschulz, line 496: Can all these keys be made "static const char xx[] = "XX";

lib/ArduinoNative/src/Arduino.cpp
```diff
@@ -130,6 +133,12 @@ static const char* PRG_ARG_RADON_ULZER_PORT_DEFAULT = "65432";
 /** Program argument default value of the configuration file. */
 static const char* PRG_ARG_CFG_FILE_DEFAULT = "config/config.json";
 
+/** Program argument default value of the Platoon ID. */
+static const char* PRG_ARG_PLATOON_ID_DEFAULT = "0";
```
nhjschulz, line 137: you can save the pointer by static const char PRG_ARG_CFG_FILE_DEFAULT[] = "config/config.json";

with existing declaration you create a pointer and a C-string with the pointer point to it.
With the proposed change you only have a C-String (4 bytes down without loosing anything :) )

lib/Service/src/SettingsHandler.cpp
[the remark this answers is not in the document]
BlueAndi: Which ones?

lib/PlatoonService/src/V2VClient.cpp
gabryelreyes: Output is published to next vehicle.

lib/Service/src/SettingsHandler.cpp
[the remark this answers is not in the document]
gabryelreyes: https://arduinojson.org/v6/error/no-matching-function-for-call-to-converttojson/

lib/PlatoonService/src/V2VClient.cpp
BlueAndi: outputTopicId is the vehicleId of the follower ... could we rename it to a more appropriate name?

lib/ArduinoNative/src/Arduino.cpp
gabryelreyes: What about something like this:
```
namespace configKeys
{
    /** Robot name */
    static const char ROBOT_NAME[] = "robotName";

    /** WiFi */
    static const char WIFI[] = "wifi";

    /** MQTT */
    static const char MQTT[] = "mqtt";

    /** Access point */
    static const char AP[] = "ap";

    /** WebServer */
    static const char WEBSERVER[] = "webServer";

    /** Platoon */
    static const char PLATOON[] = "platoon";

    /** SSID */
    static const char SSID[] = "ssid";

    /** Passphrase */
    static const char PASSWORD[] = "pswd";

    /** Host */
    static const char HOST[] = "host";

    /** Port */
    static const char PORT[] = "port";

    /** User */
    static const char USER[] = "user";

    /** Platoon ID */
    static const char PLATOON_ID[] = "platoonId";

    /** Vehicle ID */
    static const char VEHICLE_ID[] = "vehicleId";

} // namespace configKeys
```

lib/Service/src/SettingsHandler.cpp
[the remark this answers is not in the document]
BlueAndi: String is supported, see target. There seems something missing so ArduinoJson recognizes it as valid String.

lib/ArduinoNative/src/Arduino.cpp
gabryelreyes: It could be extracted into a separate file to prevent code duplication in SettingsHandler.cpp but I am not sure where such file should be located

lib/Service/src/SettingsHandler.cpp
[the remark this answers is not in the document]
gabryelreyes: String is definitely supported. I am trying to implement the fix suggested by the library [here](https://arduinojson.org/v6/error/no-matching-function-for-call-to-converttojson/), but it does not work yet. We could make a new issue for a future improvement,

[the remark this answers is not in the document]
BlueAndi:
```cpp
bool convertToJson(const String& str, JsonVariant variant)
{  
    return variant.set(str.c_str());
}
```

This is working for me.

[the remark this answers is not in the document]
gabryelreyes: Is it not _almost_ the same as assigning the values using `.c_str()` ?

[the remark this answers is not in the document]
BlueAndi: No, its not the same. We should use the ArduinoJson interface on higher levels in the same way as on target. Not starting dealing with unusal work-arounds to just get it running. Additional is supports simplicity, maintainability and readability. ;-)

lib/Service/src/SettingsHandler.cpp
```diff
@@ -280,3 +282,16 @@ SettingsHandler::~SettingsHandler()
...
+
+/**
+ * Convert a string to a json variant.
+ *
+ * @param[in]   str     String to convert.
+ * @param[out]  variant Json variant.
+ *
+ * @return If successful, it will return true otherwise false.
+ */
+bool convertToJson(const String& str, JsonVariant variant)
```
BlueAndi, line 294: Should be in ArduinoNative to be usable for all components.

BlueAndi (changes requested):

lib/Service/src/SettingsHandler.cpp
gabryelreyes: Where in ArduinoNative would it be best? Inside WString.h? `#include <ArduinoJson.h>` must also be included

lib/Service/src/SettingsHandler.cpp
```diff
@@ -57,6 +57,52 @@
  * Local Variables
  *****************************************************************************/
 
+/**
+ * Configuration keys found in the config file.
+ */
+namespace configKeys
```
BlueAndi, line 63: Put in separate header file for reuse.

BlueAndi (changes requested):

lib/ArduinoNative/src/Arduino.cpp
```diff
@@ -110,34 +110,100 @@ static const struct option LONG_OPTIONS[] = {{"help", no_argument, nullptr, 0},
...
+/** Default value of the webserver user. */
+static const char WEBSERVER_USER_DEFAULT[] = "admin";
+
+/** Default value of the webserver passphrase. */
+static const char WEBSERVER_PASSPHRASE_DEFAULT[] = "admin";
+
+/**
+ * Configuration keys found in the config file.
+ */
+namespace configKeys
```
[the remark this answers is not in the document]
gabryelreyes: Should this be with Settings (Service Folder)? That would make Arduino.cpp dependent on the Service layer

lib/ArduinoNative/src/Arduino.cpp
```diff
@@ -252,6 +318,14 @@ extern void delay(unsigned long ms)
     }
 }
 
+#ifndef UNIT_TEST
```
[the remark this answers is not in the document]
gabryelreyes: https://github.com/BlueAndi/DroidControlShip/actions/runs/7126682272/job/19405081669

Should ArduinoJson then also be a dependency of Tests?

lib/ArduinoNative/src/Arduino.cpp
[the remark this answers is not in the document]
BlueAndi: Yes, thats true. But because of key usage it makes sense. We should avoid maintaining them twice.

lib/ArduinoNative/src/Arduino.cpp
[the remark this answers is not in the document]
BlueAndi: Yes. This gives the possibility to test components, which rely on ArduinoJson too.

platformio.ini
```diff
@@ -70,7 +70,7 @@ build_flags =
     -D TARGET_NATIVE
     -D LOG_DEBUG_ENABLE=1
     -D CONFIG_LOG_SEVERITY=Logging::LOG_LEVEL_DEBUG
-    -D CONFIG_FILE_PATH="\"config/config.json\""
+    -D CONFIG_FILE_PATH=""
```
BlueAndi, line 73: Couldn't we remove it at all?

BlueAndi (changes requested):

BlueAndi (approved):

PR #65, opened 2023-12-10
Bootloader communication

gabryelreyes:


lib/HALSim/src/Device.h
```diff
@@ -110,6 +110,13 @@ class Device : public IDevice, public IDeviceNative
      */
     void enterBootloader() final;
 
+    /**
+     * Is the device in bootloader mode?
+     *
+     * @return If device is in bootloader mode, it will return true. Otherwise false.
+     */
+    bool isInBootloaderMode() const;
```
BlueAndi, line 118: final is missing

lib/HALTarget/src/Device.h
```diff
@@ -104,6 +104,13 @@ class Device : public IDevice
      */
     void enterBootloader() final;
 
+    /**
+     * Is the device in bootloader mode?
+     *
+     * @return If device is in bootloader mode, it will return true. Otherwise false.
+     */
+    bool isInBootloaderMode() const;
```
BlueAndi, line 112: final is missing

lib/HALTarget/src/USBHost.cpp
```diff
@@ -233,6 +262,32 @@ size_t USBHost::readBytes(uint8_t* buffer, size_t length)
...
+
+    if (nullptr != m_rxQueue)
+    {
+        (void)xQueueReset(m_rxQueue);
+    }
+}
+
+bool USBHost::isBootloaderModeActive() const
+{
+    return m_isBootloaderModeActive;
```
BlueAndi, line 288: How do you know that the device stays in the bootloader? It just returns that bootloader mode was requested.

BlueAndi (changes requested):

lib/HALTarget/src/USBHost.cpp
gabryelreyes: Every time the device leaves the bootloader mode, it triggers a timeout. When this happens, the flag and the connection are reset.

gabryelreyes: This is definitely the "crude" way of doing it. It should be possible to create a state machine and switch the states based on the VID and PID of the USB device. These can only be obtained during the `process()` call, that is why the `m_isBootloaderModeActive` flag is required, but it would give us a better control over which device is connected and when

BlueAndi: Ok, I understood.

BlueAndi (approved):

PR #75, opened 2024-01-23
Bugfix/logging

gabryelreyes:
Fixed nullptr check while selecting a sink and improved constructor initialization.

### DIFF
--- a/lib/Utilities/src/Logging.cpp
+++ b/lib/Utilities/src/Logging.cpp
@@ -114,14 +114,17 @@ bool Logging::selectSink(const String& name)
 
     while ((MAX_SINKS > index) && (false == status))
     {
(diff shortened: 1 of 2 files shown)


BlueAndi (approved):

lib/Utilities/src/Logging.cpp
```diff
@@ -114,14 +114,17 @@ bool Logging::selectSink(const String& name)
 
     while ((MAX_SINKS > index) && (false == status))
     {
-        if (m_sinks[index]->getName() == name)
+        if (nullptr != m_sinks[index])
```
BlueAndi, line 117: In case there is no sink registered, the index will not be updated and its an infinite loop.
Unfortuantely I saw this just one second later after merge. Thats why I reverted it again. Please fix and request again.

PR #78, opened 2024-01-24
String Class - Concat

gabryelreyes:
Enables serialization of JSON directly to String class

### DIFF
--- a/lib/ArduinoNative/src/WString.h
+++ b/lib/ArduinoNative/src/WString.h
@@ -41,6 +41,10 @@
  * Compile Switches
  *****************************************************************************/
 
+#ifndef ARDUINOJSON_ENABLE_ARDUINO_STRING
+#define ARDUINOJSON_ENABLE_ARDUINO_STRING 1
+#endif
+
 /******************************************************************************
  * Includes
  *****************************************************************************/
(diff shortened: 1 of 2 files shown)


lib/ArduinoNative/src/WString.h
```diff
@@ -648,6 +652,31 @@ class String
         }
     }
 
+    bool concat(const String& s)
+    {
+        this->operator+=(s);
```
BlueAndi, line 657: Return value?

lib/ArduinoNative/src/WString.h
```diff
@@ -648,6 +652,31 @@ class String
...
+
+    bool concat(const char* cstr)
+    {
+        if ((nullptr == cstr) || (0U == strlen(cstr)))
+        {
+            return false;
+        }
+        else
+        {
+            this->operator+=(cstr);
```
BlueAndi, line 669: Return value?

lib/ArduinoNative/src/WString.h
```diff
@@ -648,6 +652,31 @@ class String
...
+        return true;
+    }
+
+    bool concat(const char* cstr)
+    {
+        if ((nullptr == cstr) || (0U == strlen(cstr)))
+        {
+            return false;
+        }
+        else
```
BlueAndi, line 667: If the first block returns, no "else" is required.

BlueAndi (changes requested):

BlueAndi (approved):

PR #83, opened 2024-01-25
Bootloader mode confirmation

gabryelreyes:
- Bootloader mode is entered once the communication has been established.
- Removed unnecessary const attribute.

### DIFF
--- a/lib/HALInterfaces/src/IDevice.h
+++ b/lib/HALInterfaces/src/IDevice.h
@@ -101,7 +101,7 @@ public:
      *
      * @return If device is in bootloader mode, it will return true. Otherwise false.
      */
-    virtual bool isInBootloaderMode() const = 0;
+    virtual bool isInBootloaderMode() = 0;
 
 protected:
     /**
--- a/lib/HALInterfaces/src/IDevice.h
+++ b/lib/HALInterfaces/src/IDevice.h
@@ -101,7 +101,7 @@ public:
      *
      * @return If device is in bootloader mode, it will return true. Otherwise false.
      */
-    virtual bool isInBootloaderMode() = 0;
+    virtual bool isInBootloaderMode() const = 0;
 
 protected:
     /**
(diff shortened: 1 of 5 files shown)


lib/HALTarget/src/USBHost.cpp
```diff
@@ -283,9 +283,9 @@ void USBHost::reset()
     }
 }
 
-bool USBHost::isBootloaderModeActive() const
+bool USBHost::isBootloaderModeActive()
 {
-    return m_isBootloaderModeActive;
+    return m_isBootloaderModeActive && m_acm.isReady();
```
BlueAndi, line 288: Please explicit.

lib/HALInterfaces/src/IDevice.h
```diff
@@ -101,7 +101,7 @@ class IDevice
      *
      * @return If device is in bootloader mode, it will return true. Otherwise false.
      */
-    virtual bool isInBootloaderMode() const = 0;
+    virtual bool isInBootloaderMode() = 0;
```
BlueAndi, line 104: Removed because m_acm.isReady() misses the const?

BlueAndi (changes requested):

lib/HALInterfaces/src/IDevice.h
gabryelreyes: Yes! Is there a better way to do this?

BlueAndi: You can add
```mutual XXXXX m_acm;```
Fast solution, but not nice.

gabryelreyes: I am not sure what the mutual does.
Should I still make the change?

nhjschulz: mutable not mutual ?

It removes "const" qualifier from the element it is used at.

nhjschulz: I would try this, its bad if an internal used class enforces to remove const.

nhjschulz (approved):

BlueAndi (approved):

PR #84, opened 2024-01-26
Platoon: Leader App

gabryelreyes:
- Refactored Leader application.
- Added State machine.
- Added Status" SMP channel as an application heatbeat
-


lib/ConvoyLeader/src/StartupState.cpp
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    return isPending;
+}
+
+void StartupState::notifyCommandProcessed()
+{
+    if (true == m_isActive)
+    {
+        m_pendingCommandCounter++;
```
BlueAndi, line 118: Prefer post

BlueAndi (changes requested):

lib/ConvoyLeader/src/StartupState.cpp
gabryelreyes: Do you mean the pre-increment operator?

BlueAndi: yes :-)

BlueAndi (approved):

PR #85, opened 2024-01-29
Bugfix: MQTT Subscriptions Target

gabryelreyes:
Subscriptions when connected were not registered in the subscriber list


LuDubies (commented):
No findings. Accepted.

LuDubies (approved):

PR #89, opened 2024-01-31
Remote ota/development

chezlyne:


lib/ZumoOta/src/Zumo32U4Specifications.h
```diff
@@ -0,0 +1,165 @@
...
+{
+    /** The binary buffer for the response Zumo data */
+    uint8_t data[size];
+
+    /** Size of the response data in bytes */
+    uint8_t dataSize;
+};
+
+/** Namespace for specifying all used Zumo32U4 commands, return codes and expected return values */
+namespace Zumo32U4Specification
```
nhjschulz, line 70: these are definitions not declarations.  It should not be in a header as every consumer will get a copy.
If is acceptable if only included in one cpp file, but should be moved out of the header in a later patch.

nhjschulz (approved):
one recommendation which can be considered in a later patch.

PR #91, opened 2024-02-02
Remote ota/development

chezlyne:
I've just added the new logic for the next sequence.

### DIFF
--- a/lib/ZumoOta/src/BootloaderCom.cpp
+++ b/lib/ZumoOta/src/BootloaderCom.cpp
@@ -35,6 +35,7 @@
 #include "BootloaderCom.h"
 #include <Logging.h>
 #include <Board.h>
+#include <FileHandler.h>
 
 /******************************************************************************
  * Compiler Switches
@@ -101,12 +102,57 @@ private:
     static const ResponseInfo m_responses[SEQ_LENGHT]; /**< Array of response information */
 };
 
-#if 0
 class FwProgCmdProvider : public CmdProvider {
     public:
-        FwProgCmdProvider(FileHandle * fs);
-}
-#endif
+        FwProgCmdProvider(FileHandler * fs) :
+        m_index(0),
+        m_fileSystem(fs)
+    {}
+
+    virtual void reset() { m_index = 0; }
+
+    virtual bool next(const CommandInfo *& cmd, const ResponseInfo *& rsp)
+    {
+        static const uint32_t CHUNK_SIZE = 128U;
+        /*Determine the current chunk start address*/
+        uint32_t chunkStartAddress = m_index * CHUNK_SIZE;
+
+        /*Read 128 bytes from the file system*/
+        char buffer[CHUNK_SIZE];
+        String fileName = " "; /**< Bin File to be flashed.*/
+        size_t bytesRead = m_fileSystem->readFile(fileName, buffer, chunkStartAddress);
+
+        /*If no bytes were read, there are no more chunks left*/
+        if (0 == bytesRead) {
+            cmd = nullptr;
+            rsp = nullptr;
+            return false;
+        }
+        
+        /*Determine the command and response for the current chunk*/
+        uint32_t commandIndex = m_index % SEQ_LENGHT;
+        cmd = &m_cmds[commandIndex];
+        if(cmd->command == Zumo32U4Specification::WRITE_MEMORY_PAGE)
+        {
+            cmd += *buffer;
+        }
+        rsp = &m_responses[commandIndex];
+
+        /* Move to the next chunk*/
+        ++m_index;
+
+        return true;
+    }
+
+private:
+    uint32_t m_index; /**< current index in sequence.*/
+    FileHandler *m_fileSystem; /**< Instance of FileHandler.*/
+    static const CommandInfo m_cmds[]; /**< Array of command information.*/
+    static const ResponseInfo m_responses[]; /**< Array of response information.*/
+    static const uint32_t SEQ_LENGHT = 2U; /**< Length of the command sequence.*/
+
+};
+
 /******************************************************************************
  * Prototypes
  *****************************************************************************/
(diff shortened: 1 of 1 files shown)


lib/ZumoOta/src/BootloaderCom.cpp
```diff
@@ -101,12 +102,84 @@ class IntProgCmdProvider: public CmdProvider {
...
 class FwProgCmdProvider : public CmdProvider {
     public:
-        FwProgCmdProvider(FileHandle * fs);
-}
-#endif
+        /**
+         * @brief Constructor for the FwProgCmdProvider class.
+         * @param fs A pointer to the FileHandler object for handling file operations.
+         */
+        FwProgCmdProvider(FileHandler * fs) :
```
nhjschulz, line 118: Need to rethink what makesn sense as constructor parameter. 
Maybe you need the filename here and open it internally for read instead of getting a handle.

lib/ZumoOta/src/BootloaderCom.cpp
```diff
@@ -101,12 +102,84 @@ class IntProgCmdProvider: public CmdProvider {
...
+        m_fileSystem(fs)
+    {}
+
+    /**
+     * @brief Resets the command provider to its initial state.
+     *
+     * This method resets the internal index that tracks the current command
+     * in the command sequence memory.
+     */
+    virtual void reset() { m_index = 0; }
```
nhjschulz, line 129: expect more a reopen/rewind of a file here.

lib/ZumoOta/src/BootloaderCom.cpp
```diff
@@ -101,12 +102,84 @@ class IntProgCmdProvider: public CmdProvider {
...
+     * @return True if there is a next command and next response, false otherwise.
+     */
+    virtual bool next(const CommandInfo *& cmd, const ResponseInfo *& rsp)
+    {
+        static const uint32_t CHUNK_SIZE = 128U;
+        /*Determine the current chunk start address.*/
+        uint32_t chunkStartAddress = m_index * CHUNK_SIZE;
+
+        /*Read 128 bytes from the file system.*/
+        char buffer[CHUNK_SIZE];
```
nhjschulz, line 148: That cannot work. Remember the cmd processing is in the background, so you cannot put data on the stack for background operations.

lib/ZumoOta/src/BootloaderCom.cpp
```diff
@@ -101,12 +102,84 @@ class IntProgCmdProvider: public CmdProvider {
...
+        /* Move to the next chunk.*/
+        ++m_index;
+
+        return true;
+    }
+
+private:
+    uint32_t m_index; /**< current index in sequence.*/
+    FileHandler *m_fileSystem; /**< Instance of FileHandler.*/
+    static const CommandInfo m_cmds[]; /**< Array of command information.*/
```
nhjschulz, line 177: I doubt it can be const as you need to provide a dynamic address and data portion with the command

lib/ZumoOta/src/BootloaderCom.cpp
```diff
@@ -101,12 +102,84 @@ class IntProgCmdProvider: public CmdProvider {
...
+    virtual bool next(const CommandInfo *& cmd, const ResponseInfo *& rsp)
+    {
+        static const uint32_t CHUNK_SIZE = 128U;
+        /*Determine the current chunk start address.*/
+        uint32_t chunkStartAddress = m_index * CHUNK_SIZE;
+
+        /*Read 128 bytes from the file system.*/
+        char buffer[CHUNK_SIZE];
+        String fileName = " "; /**< Bin File to be flashed.*/
+        size_t bytesRead = m_fileSystem->readFile(fileName, buffer, chunkStartAddress);
```
nhjschulz, line 150: remember that you need to issue 2 commands for each chunk, but you keep reading every time.

lib/ZumoOta/src/BootloaderCom.cpp
```diff
@@ -101,12 +102,84 @@ class IntProgCmdProvider: public CmdProvider {
...
+            rsp = nullptr;
+            return false;
+        }
+        
+        /*Determine the command and response for the current chunk.*/
+        uint32_t commandIndex = m_index % SEQ_LENGHT;
+        cmd = &m_cmds[commandIndex];
+        if(cmd->command == Zumo32U4Specification::WRITE_MEMORY_PAGE)
+        {
+            cmd += *buffer;
```
nhjschulz, line 164: thsis is not doing what you want and it doesn't change command length.

lib/ZumoOta/src/BootloaderCom.cpp
```diff
@@ -101,12 +102,84 @@ class IntProgCmdProvider: public CmdProvider {
...
+        }
```
nhjschulz, line 165: where is the part that sends the address  (= chunk offset in file ) ?

nhjschulz (commented):
It goes in the right direction but still need signifcant rework to be right.

lib/ZumoOta/src/BootloaderCom.cpp
chezlyne: t will be taken care of by the next commit.

lib/ZumoOta/src/BootloaderCom.cpp
chezlyne: It will be taken care of by the next commit.

lib/ZumoOta/src/BootloaderCom.cpp
chezlyne: It will be taken care of by the next commit.

lib/ZumoOta/src/BootloaderCom.cpp
chezlyne: It will be taken care of by the next commit.

lib/ZumoOta/src/BootloaderCom.cpp
chezlyne: It will be taken care of by the next commit.

Issue #71, opened 2024-01-12
Improvement Opportunities


gabryelreyes:
- ArduinoJSON upgrade to version 7
- Network state machine using event callbacks instead of polling

gabryelreyes:
> Network/WiFi state machine using event callbacks instead of polling

The current implementation of WiFi/Network management is sufficient for our use case. Additionally, a native implementation of WiFi  has been created since this issue, which mirrors the current network management.

As there are no more issues pending, I will close this issue.

PR #92, opened 2024-02-05
V2V Communication Manager

gabryelreyes:
- Renamed V2VClient into V2VCommManager
- Added heartbeat between participants.

### DIFF
--- a/lib/PlatoonService/src/V2VCommManager.cpp
+++ b/lib/PlatoonService/src/V2VCommManager.cpp
@@ -85,7 +85,8 @@ V2VCommManager::V2VCommManager(MqttClient& mqttClient) :
     m_lastPlatoonHeartbeatTimestamp(0U),
     m_followerResponseCounter(0U),
     m_platoonHeartbeatTimer(),
-    m_vehicleHeartbeatTimeoutTimer()
+    m_vehicleHeartbeatTimeoutTimer(),
+    m_status(V2V_STATUS_NOT_INIT)
 {
 }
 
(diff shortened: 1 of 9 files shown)


lib/PlatoonService/src/V2VCommManager.cpp
```diff
@@ -0,0 +1,534 @@
...
+
+    return isSuccessful;
+}
+
+V2VCommManager::V2VStatus V2VCommManager::process(VehicleStatus status)
+{
+    m_vehicleStatus = status;
+
+    /* Is MQTT client connected? */
+    if ((true == m_mqttClient.isConnected()) && (V2V_STATUS_OK == m_v2vStatus))
```
BlueAndi, line 157: A v2v status other than OK will never heal.

lib/PlatoonService/src/V2VCommManager.cpp
```diff
@@ -0,0 +1,534 @@
...
+    if (DeserializationError::Ok != error)
+    {
+        LOG_ERROR("JSON Deserialization Error %d.", error);
+    }
+    else
+    {
+        /* Send vehicle heartbeat. */
+        JsonVariant                                 jsonTimestamp = jsonPayload["timestamp"]; /* Timestamp [ms]. */
+        StaticJsonDocument<JSON_HEARTBEAT_MAX_SIZE> heartbeatDoc;
+        String                                      heartbeatPayload;
```
BlueAndi, line 289: Scope of heartbeat* can be reduced

lib/PlatoonService/src/V2VCommManager.cpp
```diff
@@ -0,0 +1,534 @@
...
+
+        if (false == jsonTimestamp.isNull())
+        {
+            heartbeatDoc["id"]        = m_vehicleId;
+            heartbeatDoc["timestamp"] = jsonTimestamp.as<uint32_t>();
```
BlueAndi, line 294: Please add a comment why its important to send the timestamp back.

lib/PlatoonService/src/V2VCommManager.cpp
```diff
@@ -0,0 +1,534 @@
...
+                LOG_ERROR("Received heartbeat from unknown vehicle %d.", id);
+            }
+            else if (timestamp != m_lastPlatoonHeartbeatTimestamp)
+            {
+                LOG_ERROR("Received heartbeat from vehicle %d with timestamp %d, expected %d.", id, timestamp,
+                          m_lastPlatoonHeartbeatTimestamp);
+            }
+            else
+            {
+                uint8_t   followerArrayIndex = id - 1U; /*Already checked id != 0U */
```
BlueAndi, line 350: What about larger than the array? ;-)

lib/PlatoonService/src/V2VCommManager.cpp
```diff
@@ -0,0 +1,534 @@
...
+                }
+            }
+        }
+    }
+}
+
+bool V2VCommManager::setupWaypointTopics(uint8_t platoonId, uint8_t vehicleId)
+{
+    bool    isSuccessful  = false;
+    uint8_t nextVehicleId = vehicleId + 1U; /* Output is published to next vehicle. */
```
BlueAndi, line 373: Is there always a next vehicle?

lib/PlatoonService/src/V2VCommManager.cpp
```diff
@@ -0,0 +1,534 @@
...
+    }
+    else if (false == m_mqttClient.publish(m_platoonHeartbeatTopic, false, heartbeatPayload))
+    {
+        LOG_ERROR("Failed to publish MQTT message to %s.", m_platoonHeartbeatTopic.c_str());
+    }
+    else
+    {
+        LOG_DEBUG("Sent platoon heartbeat: %s", heartbeatPayload.c_str());
+        isSuccessful                    = true;
+        m_lastPlatoonHeartbeatTimestamp = timestamp;
```
BlueAndi, line 522: Comment whats the idea behind sending the timestamp.

BlueAndi (changes requested):

lib/PlatoonService/src/V2VCommManager.cpp
gabryelreyes: In the case of the last follower, it is thought of a "virtual" next vehicle. This allows for feedback to the leader, when the leader subscribes to this topic. And the platoon can be expanded without any change in the code. In this way the exception handling is only for the leader and this is done separately, otherwise the last follower would also have to get its own exception handling

lib/PlatoonService/src/V2VCommManager.cpp
gabryelreyes: Is intended until an Error handler in the application is developed

lib/PlatoonService/src/V2VCommManager.h
```diff
@@ -0,0 +1,302 @@
...
+     * @return Number of waypoints in the queue.
+     */
+    size_t getWaypointQueueSize() const;
+
+private:
+    /** Max topic length */
+    static const uint8_t MAX_TOPIC_LENGTH = 64U;
+
+    /** Number of followers. */
+    static const uint8_t NUMBER_OF_FOLLOWERS = 1U;
```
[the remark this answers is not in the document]
gabryelreyes: As platoon registering is out of scope, it is hardcoded in here. In the future it may be dynamic or passed using the config.jsonf file

lib/PlatoonService/src/V2VCommManager.h
```diff
@@ -0,0 +1,302 @@
...
+
+    /**
+     * Queue for the received waypoints.
+     * Stores pointers to the waypoints in the queue when received in the callback.
+     * The queue is emptied by the getNextWaypoint() method.
+     * The queue is filled by the targetWaypointTopicCallback() method.
+     *
+     * @tparam Waypoint*    Pointer to a Waypoint.
+     */
+    std::queue<Waypoint*> m_waypointQueue;
```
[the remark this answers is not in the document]
gabryelreyes: No static queues with max. number of elements available in Simulation. Added to [#71](https://github.com/BlueAndi/DroidControlShip/issues/71)

lib/PlatoonService/src/V2VCommManager.cpp
gabryelreyes: Also checked with `if (id > NUMBER_OF_FOLLOWERS)`

lib/PlatoonService/src/V2VCommManager.cpp
gabryelreyes: Done.

lib/PlatoonService/src/V2VCommManager.cpp
gabryelreyes: Done.

lib/PlatoonService/src/V2VCommManager.cpp
```diff
@@ -0,0 +1,534 @@
...
+    m_platoonHeartbeatTimer(),
+    m_vehicleHeartbeatTimeoutTimer(),
+    m_v2vStatus(V2V_STATUS_NOT_INIT),
+    m_vehicleStatus(),
+    m_followers{}
+{
+}
+
+V2VCommManager::~V2VCommManager()
+{
```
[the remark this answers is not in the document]
gabryelreyes: Done.

lib/PlatoonService/src/V2VCommManager.cpp
gabryelreyes: Will be fixed in the next PR. Right now it works because the leade subscribes to the next vehicle after the last follower, but this will be changed once events are introduced

BlueAndi: ok

lib/PlatoonService/src/V2VCommManager.h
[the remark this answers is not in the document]
BlueAndi: ok

BlueAndi (approved):

PR #100, opened 2024-02-12
Leader: Proximity Sensors

gabryelreyes:
- Added proximity sensors to leader and speed regulation based on distance to object in the way.
- Added V2V Emergency event
- Check V2V and RU states once. Prevent continous calling of error state

### DIFF
--- a/lib/ConvoyFollower/src/App.cpp
+++ b/lib/ConvoyFollower/src/App.cpp
@@ -206,7 +206,7 @@ void App::loop()
     {
         /* Log and Handle Board processing error */
         LOG_FATAL("HAL process failed.");
-        fatalErrorHandler();
+        setErrorState();
     }
 
     /* Process SerialMuxProt. */
@@ -227,7 +227,11 @@ void App::loop()
 
 void App::setErrorState()
 {
-    m_systemStateMachine.setState(&ErrorState::getInstance());
+    if (&ErrorState::getInstance() != m_systemStateMachine.getState())
+    {
+        m_systemStateMachine.setState(&ErrorState::getInstance());
+        m_v2vCommManager.triggerEmergencyStop();
+    }
 }
 
 void App::systemStatusCallback(SMPChannelPayload::Status status)
@@ -458,6 +462,7 @@ void App::processV2VCommunication()
 
     v2vStatus = m_v2vCommManager.process(vehicleStatus);
 
+    /* Process V2V status. */
     switch (v2vStatus)
     {
     case V2VCommManager::V2V_STATUS_OK:
@@ -469,9 +474,15 @@ void App::processV2VCommunication()
         break;
 
     case V2VCommManager::V2V_STATUS_LOST_FOLLOWER:
+        /* Fallthrough */
     case V2VCommManager::V2V_STATUS_FOLLOWER_ERROR:
+        /* Fallthrough */
     case V2VCommManager::V2V_STATUS_OLD_WAYPOINT:
         LOG_ERROR("Follower Error");
+        setErrorState();
+        break;
+
+    case V2VCommManager::V2V_STATUS_EMERGENCY:
         setErrorState();
         break;
 
--- a/lib/ConvoyFollower/src/App.cpp
+++ b/lib/ConvoyFollower/src/App.cpp
@@ -236,20 +236,26 @@ void App::setErrorState()
 
 void App::systemStatusCallback(SMPChannelPayload::Status status)
 {
-    switch (status)
+    if (m_lastRUStatus != status)
     {
-    case SMPChannelPayload::STATUS_FLAG_OK:
-        /* Nothing to do. All good. */
-        break;
+        /* Save last status. */
+        m_lastRUStatus = status;
 
-    case SMPChannelPayload::STATUS_FLAG_ERROR:
-        LOG_DEBUG("RU Status ERROR.");
-        setErrorState();
-        m_statusTimeoutTimer.stop();
-        break;
+        switch (status)
+        {
+        case SMPChannelPayload::STATUS_FLAG_OK:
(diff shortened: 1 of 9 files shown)


lib/ConvoyFollower/src/App.cpp
```diff
@@ -227,25 +227,35 @@ void App::loop()
...
 
-    case SMPChannelPayload::STATUS_FLAG_ERROR:
-        LOG_DEBUG("RU Status ERROR.");
-        setErrorState();
-        m_statusTimeoutTimer.stop();
-        break;
+        switch (status)
+        {
+        case SMPChannelPayload::STATUS_FLAG_OK:
+            /* Nothing to do. All good. */
```
BlueAndi, line 247: If there is only one good case and all others are errors, you can use a simple if/else

lib/ConvoyFollower/src/App.cpp
```diff
@@ -458,30 +468,43 @@ void App::processV2VCommunication()
...
+            /* Fallthrough */
+        case V2VCommManager::V2V_STATUS_OLD_WAYPOINT:
+            LOG_ERROR("Follower Error");
+            setErrorState();
+            break;
 
-    default:
-        break;
+        case V2VCommManager::V2V_STATUS_EMERGENCY:
+            setErrorState();
```
BlueAndi, line 497: What is the difference between error and emergency?

lib/ConvoyFollower/src/App.h
```diff
@@ -78,7 +78,9 @@ class App
         m_commandTimer(),
         m_motorSpeedTimer(),
         m_statusTimer(),
-        m_statusTimeoutTimer()
+        m_statusTimeoutTimer(),
+        m_lastV2VStatus(V2VCommManager::V2VStatus::V2V_STATUS_OK),
```
BlueAndi, line 82: Shouldn't it be something like not init?

lib/ConvoyLeader/src/App.cpp
```diff
@@ -236,25 +236,35 @@ void App::setLatestVehicleData(const Waypoint& waypoint)
...
 
-    case SMPChannelPayload::STATUS_FLAG_ERROR:
-        LOG_ERROR("RU Status ERROR.");
-        setErrorState();
-        m_statusTimeoutTimer.stop();
-        break;
+        switch (status)
+        {
+        case SMPChannelPayload::STATUS_FLAG_OK:
+            /* Nothing to do. All good. */
```
BlueAndi, line 256: Same point as before.

lib/ConvoyLeader/src/App.h
```diff
@@ -79,7 +79,9 @@ class App
         m_commandTimer(),
         m_motorSpeedTimer(),
         m_statusTimer(),
-        m_statusTimeoutTimer()
+        m_statusTimeoutTimer(),
+        m_lastV2VStatus(V2VCommManager::V2VStatus::V2V_STATUS_OK),
```
BlueAndi, line 83: not init?

lib/ConvoyLeader/src/DrivingState.cpp
```diff
@@ -73,12 +73,24 @@ void DrivingState::process(StateMachine& sm)
...
 
+        /* Limit speed based on the proximity sensors. */
+        /* m_vehicleData.proximity goes from 0 to NUM_PROXIMITY_SENSOR_RANGES */
+        maxPossibleSpeed = m_maxMotorSpeed - (m_vehicleData.proximity * m_rangeFactor);
+
+        /* Ensure that the robot stops when an obstable is too near. */
+        if (NUM_PROXIMITY_SENSOR_RANGES == m_vehicleData.proximity)
+        {
+            maxPossibleSpeed = 0;
+        }
```
BlueAndi, line 90: calculate only in else part

lib/ConvoyLeader/src/DrivingState.cpp
```diff
@@ -90,6 +102,7 @@ void DrivingState::exit()
 void DrivingState::setMaxMotorSpeed(int16_t maxSpeed)
 {
     m_maxMotorSpeed = maxSpeed;
+    m_rangeFactor   = m_maxMotorSpeed / NUM_PROXIMITY_SENSOR_RANGES;
```
BlueAndi, line 105: m_rangeFactor can be used just in place without member variable or?

lib/ConvoyLeader/src/DrivingState.cpp
```diff
@@ -73,12 +73,24 @@ void DrivingState::process(StateMachine& sm)
...
     {
+        int16_t maxPossibleSpeed = m_maxMotorSpeed;
+
         /* TODO: Check limits. */
 
         /* TODO: Check follower feedback. Calculate platoon length and react accordingly */
 
+        /* Limit speed based on the proximity sensors. */
+        /* m_vehicleData.proximity goes from 0 to NUM_PROXIMITY_SENSOR_RANGES */
+        maxPossibleSpeed = m_maxMotorSpeed - (m_vehicleData.proximity * m_rangeFactor);
```
BlueAndi, line 84: Is there any processing time considered, e.g. from detection time to reaction time 10 ms or more? Or just a proof-of-concept?

BlueAndi (changes requested):

lib/ConvoyFollower/src/App.cpp
gabryelreyes: Error is unexpected error. Emergency is purposefully triggered.

BlueAndi: Shouldn't it be handled then different?

BlueAndi (changes requested):

BlueAndi (approved):

PR #101, opened 2024-02-12
Remote ota/development

chezlyne:


lib/ZumoOta/src/FileManager.cpp
```diff
@@ -80,3 +81,32 @@ bool FileManager::init()
     }
     return true;
 }
+
+size_t FileManager::read128Bytes(const char* firmwareName, uint8_t* buffer)
```
nhjschulz, line 85: Thats a rather strange convetion to have a readxXbytes function. It is unflexible for no gain.  User readBytes( len ...) instead and call it with 128 from the flash prog.

lib/ZumoOta/src/BootloaderCom.cpp
```diff
@@ -97,26 +102,122 @@ class IntProgCmdProvider: public CmdProvider {
...
+
+        if(cmd->command == Zumo32U4Specification::SET_MEMORY_ADDR)
+        {
+            cmd->commandsize = FinalLength;
+        }
+        else if(cmd->command == Zumo32U4Specification::WRITE_MEMORY_PAGE)
+        {
+            size_t bytesRead = m_fileManager.read128Bytes(m_fileName.c_str(), m_buffer);
+            /*If no bytes were read, there are no more chunks left.*/
+            if (0 == bytesRead)
```
nhjschulz, line 161: What do you do if your firmware ist 129 bytes long ?

You have to handle the fact that the last block is in between 0 and 127 tail bytes.
In this case you need to fill the gap to the missing blocksize with a pattern, usually 0xFF or 0x00,
The value is what the flash reads after an erase, to avoid unnecessary programming of the bits.

lib/ZumoOta/src/BootloaderCom.cpp
```diff
@@ -97,26 +102,122 @@ class IntProgCmdProvider: public CmdProvider {
...
+        rsp = &m_responses[commandIndex];
+
+        /* Move to the next chunk.*/
+        ++m_index;
+
+        return true;
+    }
+
+    FileManager m_fileManager; /**< Instance of Filemanger class.*/
+    const size_t FinalLength = 3;/**< Final Size of the updated Command.*/
```
nhjschulz, line 181: es you use this only in one place, I don't see the value of hacing this constant. Just explain what you assign at line 155

lib/ZumoOta/src/BootloaderCom.cpp
```diff
@@ -97,26 +102,122 @@ class IntProgCmdProvider: public CmdProvider {
...
+
+private:
+    uint8_t m_index; /**< current index in sequence.*/
+    static  CommandInfo m_cmds[]; /**< Array of command information.*/
+    static  ResponseInfo m_responses[]; /**< Array of response information.*/
+    std::string m_fileName = "/zumo_firmware.bin"; /**< Name of the file to be read.*/
```
nhjschulz, line 187: why a string if it is just a character array constant ?
Does this project ust c++ strings at all ?

lib/ZumoOta/src/BootloaderCom.cpp
```diff
@@ -97,26 +102,122 @@ class IntProgCmdProvider: public CmdProvider {
...
+    static const uint32_t SEQ_LENGHT = 2U; /**< Length of the command sequence.*/
```
nhjschulz, line 188: I don't understand what length this is. The programming is for sure longer then 2 commands.

It seems just to be a toggle between set_addr and writ_page.  So why not be explicit with somthing like 

enum ProgCmds  {FLSPRG_SET_ADDR, FLSPRG_WRITE_BLOCK};
ProgrCmds nextCmd = FLSPRG_SET_ADDR;

instead of the misleading SEQ_LENGTH. This would be way better to understand.

lib/ZumoOta/src/BootloaderCom.cpp
```diff
@@ -227,13 +341,63 @@ bool BootloaderCom::process()
             m_waitingForResponse = false;
             if(m_cmdProvider->next(m_currentCommand, m_currentResponse))
             {
-                if(true == m_flashManager.sendCommand(
-                    m_currentCommand->command,
-                    m_currentCommand->commandsize))
+                if(m_currentCommand->command == Zumo32U4Specification::SET_MEMORY_ADDR)
```
nhjschulz, line 344: The block from here to line 392 belongs into the cmd provider for the flash process not here.  this function executs cmds from a provider, it does nut build/modify them. That is the duty of the provider not the executer (seperation of concerns).

lib/ZumoOta/src/BootloaderCom.cpp
```diff
@@ -477,8 +641,45 @@ bool BootloaderCom::compareExpectedAndReceivedResponse(const uint8_t command[],
     return true;
 }
 
+ void BootloaderCom:: updateCommand(uint8_t newCommand[])
```
nhjschulz, line 644: generic function name that is only applicable for set_addr. 
And it doesn't belong here, but in the provider for programming,

lib/ZumoOta/src/BootloaderCom.cpp
```diff
@@ -477,8 +641,45 @@ bool BootloaderCom::compareExpectedAndReceivedResponse(const uint8_t command[],
...
+        m_currWriteMemAddr += 64U;
+    }
+    else
+    {
+        LOG_ERROR("newCommand is a nullptr!");
+    }
 
+}
+
+void BootloaderCom:: update2Command(uint8_t newCommand[], uint8_t* buffer, size_t bufferSize)
```
nhjschulz, line 662: same as former comment.

lib/ZumoOta/src/BootloaderCom.cpp
```diff
@@ -75,8 +79,9 @@ class IntProgCmdProvider: public CmdProvider {
      * @param rsp Reference to a pointer to the next response.
      * @return True if there is a next command and next response, false otherwise.
      */
-    virtual bool next( const CommandInfo *& cmd,  const ResponseInfo *& rsp)
+    virtual bool next(  CommandInfo *& cmd,   ResponseInfo *& rsp)
```
nhjschulz, line 82: the const was there before by purpose. See comment at line 344 below. the process() routine shall not modify commands, the provider need to feed with with ready to use ones.

lib/ZumoOta/src/BootloaderCom.cpp
```diff
@@ -227,13 +341,63 @@ bool BootloaderCom::process()
...
             {
-                if(true == m_flashManager.sendCommand(
-                    m_currentCommand->command,
-                    m_currentCommand->commandsize))
+                if(m_currentCommand->command == Zumo32U4Specification::SET_MEMORY_ADDR)
+                {
+                    size_t newSize = m_currentCommand->commandsize;
+                    uint8_t newCommand[newSize];
+                    std::memcpy(newCommand, m_currentCommand->command, m_currentCommand->commandsize);
+                    // Rufen Sie die Funktion zur Aktualisierung des Befehls auf
```
nhjschulz, line 349: convention is english language

lib/ZumoOta/src/FileManager.h
```diff
@@ -15,7 +15,7 @@
 /******************************************************************************
  * Includes
  *****************************************************************************/
-
+#include <Arduino.h>
```
nhjschulz, line 18: why ??

nhjschulz (commented):
Several issues to rework.

lib/ZumoOta/src/BootloaderCom.cpp
```diff
@@ -97,26 +102,122 @@ class IntProgCmdProvider: public CmdProvider {
 private:
     static const uint32_t SEQ_LENGHT = 12; /**< Length of the command sequence */
```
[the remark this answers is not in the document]
chezlyne: It will be explained in the next commit.

lib/ZumoOta/src/FileManager.cpp
chezlyne: It will be adjusted in the next commit.

lib/ZumoOta/src/BootloaderCom.cpp
chezlyne: It will be adjusted in the next commit.

lib/ZumoOta/src/BootloaderCom.cpp
chezlyne: It will be adjusted in the next commit.

lib/ZumoOta/src/BootloaderCom.cpp
chezlyne: It will be adjusted in the next commit.

lib/ZumoOta/src/BootloaderCom.cpp
chezlyne: It will be adjusted in the next commit.

lib/ZumoOta/src/BootloaderCom.cpp
chezlyne: It will be adjusted in the next commit.

lib/ZumoOta/src/BootloaderCom.cpp
chezlyne: It will be adjusted in the next commit.

lib/ZumoOta/src/BootloaderCom.cpp
chezlyne: It will be adjusted in the next commit.

lib/ZumoOta/src/FileManager.h
chezlyne: It will be removed in the next commit.

PR #108, opened 2024-03-18
Sensor fusion/ End Line Detection Removal 

jkerpe:
- Removed End Line Detection Channel
- Only perform the Update Step of the EKF if the Odometry values have changed

### DIFF
--- a/lib/SensorFusion/src/App.cpp
+++ b/lib/SensorFusion/src/App.cpp
@@ -60,7 +60,6 @@
  *****************************************************************************/
 
 static void App_sensorChannelCallback(const uint8_t* payload, const uint8_t payloadSize, void* userData);
-static void App_endlineChannelCallback(const uint8_t* payload, const uint8_t payloadSize, void* userData);
 
 /******************************************************************************
  * Local Variables
@@ -162,7 +161,6 @@ void App::setup()
 
             /* Setup SerialMuxProt Channels */
             m_smpServer.subscribeToChannel(SENSORDATA_CHANNEL_NAME, App_sensorChannelCallback);
-            m_smpServer.subscribeToChannel(ENDLINE_CHANNEL_NAME, App_endlineChannelCallback);
 
             if (false == m_mqttClient.init())
             {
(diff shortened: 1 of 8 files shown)


lib/SensorFusion/src/ExtendedKalmanFilter.h
```diff
@@ -95,17 +95,20 @@ class ExtendedKalmanFilter : public IKalmanFilter
...
 
     /**
-     * Prediction of the covariance and the state of the Extended Kalman Filter.
+     * Prediction of the covariance and the state of the Kalman Filter.
+     * 
      * @param[in] timeStep Measured Time Step in ms.
+     * @param[in] kalmanParameter   Input Parameters for the Kalman Filter
      */
-    void predictionStep(const uint16_t timeStep) final;
+    virtual void predictionStep(const uint16_t timeStep, KalmanParameter& kalmanParameter) final;
```
gabryelreyes, line 103: The parameters can change between each step, or why do you need to pass them each time?

lib/SensorFusion/src/App.cpp
```diff
@@ -231,9 +229,9 @@ void App::publishSensorFusionPosition()
 
     /* Write newest Sensor Fusion Data in JSON String */
     IKalmanFilter::PositionData currentPosition = m_sensorFusion.getLatestPosition();
-    payloadJson["positionX"]                    = currentPosition.positionX;
-    payloadJson["positionY"]                    = currentPosition.positionY;
-    payloadJson["angle"]                        = currentPosition.angle;
+    payloadJson["positionX"]                    = static_cast<int32_t>(currentPosition.positionX);
```
gabryelreyes, line 232: Add 0.5F to round correctly from float

gabryelreyes (changes requested):

lib/SensorFusion/src/ExtendedKalmanFilter.h
jkerpe: Yes, these are the Measurement Values from the IMU and the Odometry.

gabryelreyes (approved):

PR #109, opened 2024-03-19
Added tests for platoon utils

gabryelreyes:
- Used tests to fix some implementation issues in the equivalent heading calculation.
- Added a HALTest in order to test libraries that require HAL includes, even if these are not used.

### DIFF
--- a/lib/HALTest/src/FileHandler.cpp
+++ b/lib/HALTest/src/FileHandler.cpp
@@ -1,0 +1,84 @@
+/* MIT License
+ *
+ * Copyright (c) 2023 - 2024 Andreas Merkle <web@blue-andi.de>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+/*******************************************************************************
+    DESCRIPTION
+*******************************************************************************/
+/**
+ *  @brief  File Handler
+ *  @author Gabryel Reyes <gabryelrdiaz@gmail.com>
+ */
+
+/******************************************************************************
+ * Includes
+ *****************************************************************************/
+
+#include "FileHandler.h"
+#include <stdio.h>
+#include <Logging.h>
+
+/******************************************************************************
+ * Macros
+ *****************************************************************************/
+
+/******************************************************************************
+ * Prototypes
+ *****************************************************************************/
+
+/******************************************************************************
+ * Local Variables
+ *****************************************************************************/
+
+/******************************************************************************
+ * Public Methods
+ *****************************************************************************/
+
+FileHandler::FileHandler()
+{
+}
+
+FileHandler::~FileHandler()
+{
+}
+
+size_t FileHandler::readFile(const String& fileName, char* outBuffer, const uint32_t maxBufferSize) const
+{
+    size_t readBytes = 0U;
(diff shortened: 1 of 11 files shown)


lib/HALTest/src/FileHandler.cpp
```diff
@@ -0,0 +1,84 @@
...
+}
+
+FileHandler::~FileHandler()
+{
+}
+
+size_t FileHandler::readFile(const String& fileName, char* outBuffer, const uint32_t maxBufferSize) const
+{
+    size_t readBytes = 0U;
+    LOG_WARNING("FileHandler::readFile not implemented for test environment.");
```
BlueAndi, line 67: Please use NOT_USED() macro.

lib/HALTest/src/MqttClient.cpp
```diff
@@ -0,0 +1,143 @@
...
+    UTIL_NOT_USED(topic);
+    UTIL_NOT_USED(useClientIdAsBaseTopic);
+    UTIL_NOT_USED(message);
+    LOG_WARNING("MQTT client not implemented for test environment.");
+    return false;
+}
+
+bool MqttClient::subscribe(const String& topic, const bool useClientIdAsBaseTopic, TopicCallback callback)
+{
+    LOG_WARNING("MQTT client not implemented for test environment.");
```
BlueAndi, line 118: Please use NOT_USED() macro.

lib/HALTest/src/MqttClient.h
```diff
@@ -0,0 +1,149 @@
...
+    bool subscribe(const String& topic, const bool useClientIdAsBaseTopic, TopicCallback callback) final;
+
+    /**
+     * Unsubscribes from a topic.
+     *
+     * @param[in] topic                     Topic to unsubscribe from.
+     * @param[in] useClientIdAsBaseTopic    If true, the client ID is used as the base (prefix) of the topic.
+     */
+    void unsubscribe(const String& topic, const bool useClientIdAsBaseTopic) final;
+};
```
BlueAndi, line 142: Can the original one be copied too?

lib/PlatoonService/src/PlatoonUtils.h
```diff
@@ -67,6 +67,7 @@ namespace PlatoonUtils
 
     /**
      * Calculate the heading at which a target waypoint is located relative to other waypoint.
+     * The result is the heading in mrad. The result is always positive, between 0 and 2*PI.
```
BlueAndi, line 70: Means between 0 < x < 2PI or [0; 2PI] ?

lib/PlatoonService/src/PlatoonUtils.cpp
```diff
@@ -117,13 +117,27 @@ bool PlatoonUtils::areWaypointsEqual(const Waypoint& waypoint1, const Waypoint&
 
 int32_t PlatoonUtils::calculateEquivalentHeading(int32_t targetHeading, int32_t referenceHeading)
 {
-    int32_t equivalentHeading = targetHeading;
+    int32_t equivalentHeading = targetHeading % FP_2PI();
+
+    if (FP_2PI() == referenceHeading)
```
BlueAndi, line 122: 2PI and -2PI are the same as 0, so here I think the referenceHeading should be set to 0, because the equivalentHeading is already 0 by modulo.

BlueAndi (changes requested):

lib/PlatoonService/src/PlatoonUtils.cpp
gabryelreyes: Agreed, but this covers the case in which the measured heading has not performed a modulo operation. If this is the case, reseting the angle to 0 would cause it to turn the wrong way. The reference heading cannot be modified/set to 0 with modulo, therefore this case must be handled.

lib/HALTest/src/MqttClient.h
gabryelreyes: The MQTTClient of Sim? Or which original do you mean?

lib/PlatoonService/src/PlatoonUtils.h
```diff
@@ -67,6 +67,7 @@ namespace PlatoonUtils
 
     /**
      * Calculate the heading at which a target waypoint is located relative to other waypoint.
+     * The result is the heading in mrad. The result range is [-PI, PI].
```
BlueAndi, line 70: Its the question for all other functions with angles too ... what is the expected range.

BlueAndi (changes requested):

lib/PlatoonService/src/PlatoonUtils.cpp
BlueAndi: I don't think you find it in a different range, because 0..2PI is the whole range, as well as -PI...PI contains it. But we should use internally always the same range to not confuse ourself. :-)

gabryelreyes: One example is when reference is 2PI and target it PI/2. The closest equivalent angle is reached counter-clockwise. However as the angle PI/2 < 2PI, the robot will turn clockwise all the way to PI/2 instead. This is what the different ranges here are trying to prevent

lib/HALTest/src/MqttClient.h
BlueAndi: The one you address with this stubbed one.

lib/PlatoonService/src/PlatoonUtils.cpp
BlueAndi: I am talking about the ranges of the input values only. :-)
If you change the range inside the function, please comment why and to which range. This makes it more understandable.

BlueAndi (changes requested):

lib/HALTest/src/MqttClient.h
gabryelreyes: Sorry I dont understand what you mean

BlueAndi (approved):

PR #118, opened 2024-05-17
Introduction of the webots serial driver for RU communication

BlueAndi:


lib/APPTest/App.cpp
gabryelreyes, line 1: Please check the file structure of the other applications and repeat it here. Source and header files should be inside a src folder

platformio.ini
```diff
@@ -72,15 +72,26 @@ build_flags =
...
     HALInterfaces
     ArduinoNative
     HALSim
+    Webots
 lib_ignore =
     HALTarget
     HALTest
 extra_scripts =
+    ./scripts/webots_launcher.py
     pre:./scripts/add_os_specific_build_flags.py
```
gabryelreyes, line 89: Script `scripts\add_os_specific_build_flags.py` can now be removed, as it is only necessary for socket connection

scripts/add_os_specific_build_flags.py
gabryelreyes, line 1: No longer necessary

gabryelreyes (changes requested):

lib/APPTest/App.cpp
BlueAndi: ok

platformio.ini
BlueAndi: ok

scripts/add_os_specific_build_flags.py
BlueAndi: ok

gabryelreyes (approved):

PR #120, opened 2024-05-22
Native Main

gabryelreyes:
Separated MainNative and MainTestNative from ArduinoNative

### DIFF
--- a/lib/ArduinoNative/src/Arduino.h
+++ b/lib/ArduinoNative/src/Arduino.h
@@ -90,7 +90,7 @@ namespace Arduino
      */
     void loop();
 
-} // namespace Arduino
+} /* namespace Arduino */
 
 /**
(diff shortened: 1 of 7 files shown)


lib/ArduinoNative/src/Arduino.h
```diff
@@ -62,10 +62,36 @@
...
+     */
+    void setup(GetSystemTick getSystemTickFunc, SystemDelay systemDelayFunc);
+
+    /**
+     * Loop shall be called periodically and handles Arduino library
+     * functionality and call the application loop() function.
+     */
+    void loop();
+
+} // namespace Arduino
```
BlueAndi, line 93: Replace // with /* */

BlueAndi (changes requested):

BlueAndi (approved):

PR #128, opened 2024-08-01
Added orientation to GPS using Compass

gabryelreyes:


lib/HALSim/src/Gps.cpp
```diff
@@ -94,6 +97,36 @@ bool Gps::getPosition(int32_t& xPos, int32_t& yPos)
...
+
+    if (nullptr != m_compass)
+    {
+        /* Get orientation in radians from Webots. */
+        const double* orientationVector = m_compass->getValues();
+
+        if (nullptr != orientationVector)
+        {
+            double xVector = orientationVector[0];
+            double yVector = orientationVector[1];
```
jkerpe, line 112: The names `xVector` and `yVector` are confusing --> sound multi-dimensional

gabryelreyes: It is multi-dimensional. X and Y directions :D

jkerpe: `xVector` and `yVector` are one-dimensional variables, but the `vector` inside the name make them sound multi-dimensional --> maybe something with `component` or `direction`?

lib/HALSim/src/Gps.cpp
```diff
@@ -55,17 +58,17 @@
  * Local Variables
  *****************************************************************************/
 
-/** Index of the x-coordinate in the position vector. */
-static const size_t GPS_POSITION_X_INDEX = 0U;
+/** Index of the x-component in the a vector. */
```
jkerpe, line 61: "in **the a** vector"

lib/HALSim/src/Gps.cpp
```diff
@@ -55,17 +58,17 @@
...
+static const size_t X_COMPONENT_INDEX = 0U;
 
-/** Index of the y-coordinate in the position vector. */
-static const size_t GPS_POSITION_Y_INDEX = 1U;
+/** Index of the y-component in the a vector. */
```
jkerpe, line 64: "in **the a** vector"

jkerpe (approved):

PR #140, opened 2024-08-07
Fix simulation under Linux

gabryelreyes:
Added fixes in include statements and OS-dependent APIs


lib/MainNative/src/main.cpp
```diff
@@ -49,6 +49,12 @@
  * Macros
  *****************************************************************************/
 
+#ifdef _WIN32
+#define MKDIR(dir) mkdir(dir) /**< Make new directory in Windows OS. Does not accept permissions argument. */
```
nhjschulz, line 53: This is a c++ world, why not 
```
inline bool createDir(const char * path)
{
   int retval = -1;
#ifdef _WIN32
    retval =mkdir(dir) /**< Make new directory in Windows OS. Does not accept permissions argument. */
#else
    retval  =mkdir(dir, 0755) /**< Make new directory in Linux OS. Uses 0755 as default permissions. */
#endif
  return (0 == retval );
}
```

gabryelreyes: I guess that is also a possibility. What benefits does it have over the macro?

nhjschulz: MISRA: Rule 19.7 (advisory): A function should be used in preference to a function-like macro
other: https://wiki.sei.cmu.edu/confluence/display/c/PRE00-C.+Prefer+inline+or+static+functions+to+function-like+macros

https://isocpp.org/wiki/faq/inline-functions: Why should I use inline functions instead of plain old #define macros?
Webots CS375 Avoid defining macros at all costs.

nhjschulz (approved):

Issue #114, opened 2024-04-09
Check speed units


gabryelreyes:
RU sends calculates speed in steps/s, however DCS should work with mm/s.

Check if this is correctly implemented and fix otherwise.

BlueAndi:
This is was by intention to reduce load on RU side. Its not a bug.

PR #150, opened 2024-08-14
Turtle behavior

gabryelreyes:


lib/APPTurtle/src/App.cpp
```diff
@@ -234,6 +220,126 @@ void App::fatalErrorHandler()
...
+    /* Check for new data. */
+    if (true == m_isNewTurtleSpeedSetpoint)
+    {
+        m_isNewTurtleSpeedSetpoint = false;
+        LOG_DEBUG("New turtle speed setpoint received. %f", m_turtleSpeedSetpoint.linear.x);
+        SpeedData payload;
+
+        if (m_turtleSpeedSetpoint.linear.x > 0.0F)
+        {
+            payload.left  = 3000;
```
nhjschulz, line 304: csn you explain the number value and why you not consider what x or z is ?

nhjschulz (changes requested):

lib/APPTurtle/src/App.cpp
gabryelreyes: Will add some comments, just for this PR. A rework of the app and a state machine are probably necessary to retrieve the maximum motor speed and adjust the setpoint accordingly to take the input of x and z into account

gabryelreyes:
https://github.com/BlueAndi/DroidControlShip/issues/114 must be fixed before any scaling or use of the incoming data can be properly used.

nhjschulz (approved):

PR #153, opened 2024-08-16
Use mm/s

### LINKED ISSUE #114: Check speed units

gabryelreyes:
Use mm/s instead of steps/s
Allows for the counterpart to use SI units instead of steps.
Counterpart of https://github.com/BlueAndi/RadonUlzer/pull/146

Closes #114

### DIFF
--- a/lib/ConvoyFollower/src/App.cpp
+++ b/lib/ConvoyFollower/src/App.cpp
@@ -388,8 +388,8 @@ void App::processPeriodicTasks()
 
     if ((true == m_motorSpeedTimer.isTimeout()) && (true == m_smpServer.isSynced()))
     {
-        int16_t   leftSpeed  = 0;
-        int16_t   rightSpeed = 0;
+        int32_t   leftSpeed  = 0;
+        int32_t   rightSpeed = 0;
         SpeedData payload;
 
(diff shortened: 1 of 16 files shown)


nhjschulz:
Would have been a good oportunity to introduce a user type for speed, something like typedef uint32_t velocity_t.

BlueAndi:
The change has influence on the PIDs, I only saw one change there. What about calculations at all?

gabryelreyes:
I tested the simulation and this was the only change required for the stability of the platoon

BlueAndi:
> I tested the simulation and this was the only change required for the stability of the platoon

A stability issue means there is a bug or bad parameters. There is no influence just by unit change on stability. ;-)
Anyway you can merge, but the stability issue might still be included and just damped.

BlueAndi (approved):

gabryelreyes:
Not even an issue when the maximum value possible changes? Then you may be right, and there may still be an issue

BlueAndi:
> Not even an issue when the maximum value possible changes?

No, there is always a max. speed by electronic + mechanic. Every higher value is limited on RU side. Seems for me the problem is on DCS side, that it assumes the higher speed will work, but it doesn't.

PR #156, opened 2024-08-19
Turtle Speeds

gabryelreyes:


lib/APPTurtle/src/App.cpp
```diff
@@ -274,10 +274,9 @@ bool App::setupSerialMuxProtServer()
...
     m_serialMuxProtChannelIdStatus = m_smpServer.createChannel(STATUS_CHANNEL_NAME, STATUS_CHANNEL_DLC);
-    m_serialMuxProtChannelIdMotorSpeeds =
-        m_smpServer.createChannel(SPEED_SETPOINT_CHANNEL_NAME, SPEED_SETPOINT_CHANNEL_DLC);
+    m_serialMuxProtChannelIdTurtle = m_smpServer.createChannel(TURTLE_CHANNEL_NAME, TURTLE_CHANNEL_DLC);
 
-    if ((0U == m_serialMuxProtChannelIdStatus) || (0U == m_serialMuxProtChannelIdMotorSpeeds))
+    if ((0U == m_serialMuxProtChannelIdStatus) || (0U == m_serialMuxProtChannelIdTurtle))
     {
         LOG_ERROR("Failed to create SerialMuxProt channels.");
     }
```
nhjschulz, line 282: Isn't this fatal ? I wonder why this is just a log but no SW reaction. What happens later on this error happend ?
Shouldn't we go into a HALT with error indication (like RED LED flash and RU reset pin trigger).

lib/APPTurtle/src/App.cpp
```diff
@@ -295,57 +294,38 @@ void App::handleTurtle()
...
-            payload.right = MOTOR_MAX_SPEED;
-        }
-        /* Turn right. */
-        else if (m_turtleSpeedSetpoint.angular.z < 0.0F)
-        {
-            payload.left  = MOTOR_MAX_SPEED;
-            payload.right = -MOTOR_MAX_SPEED;
-        }
+        TurtleSpeed payload;
+        int32_t     linearSpeed  = m_turtleSpeedSetpoint.linear.x * 1000;  /* Linear speed in mm/s */
```
nhjschulz, line 298: Scale is a magic number, better use a constant or inline to make it clear. From code it is not obvious what setpoint.linear is.

Example:
/**
   * Scale linear speed from twist message (meter/s) to milimeter/s
   */
inline  int32_t scaleTwistSpeed(int32_t speed)
{
   return meterPerSec *1000;  /* to mm/s *
}

lib/APPTurtle/src/App.cpp
```diff
@@ -295,57 +294,38 @@ void App::handleTurtle()
...
+        int32_t     angularSpeed = m_turtleSpeedSetpoint.angular.z * 1000; /* Angular speed in mrad/s */
```
nhjschulz, line 299: same as above, use inline or a constant instead of 1000.

lib/APPTurtle/src/App.cpp
```diff
@@ -295,57 +294,38 @@ void App::handleTurtle()
...
+
+        payload.linearCenter = linearSpeed;
+        payload.angular      = angularSpeed;
+
+        LOG_DEBUG("Linear speed: %d mm/s, Angular speed: %d mrad/s", payload.linearCenter, payload.angular);
 
-        if (false == m_smpServer.sendData(m_serialMuxProtChannelIdMotorSpeeds, &payload, sizeof(payload)))
+        if (false == m_smpServer.sendData(m_serialMuxProtChannelIdTurtle, &payload, sizeof(payload)))
         {
             LOG_WARNING("Failed to send motor speeds to RU.");
```
nhjschulz, line 308: Again: Error Handling in this case ? Stop RU via reset, RED led flash ?

nhjschulz (changes requested):

lib/APPTurtle/src/App.cpp
gabryelreyes: It is in fact fatal. Please see line 136

lib/APPTurtle/src/App.cpp
gabryelreyes: I do not believe this is a fatal error and the system should be stopped. RU already stops if communication is lost, so it is not a safety concern. And stopping the system would prevent any chance at retrying at a later time

nhjschulz: ok

nhjschulz (approved):

PR #160, opened 2024-08-29
Bugfix: Added manual simulation sime step to enable busy waits in ROS executor spin

hoeftjch:
The ROS session recv_message requires busy waits for timeouts. The simulation only progresses in time after a completed application loop. Until further changes the simulation time will be progressed while waiting for the SimpleTimer timeout loop in custom_transport_read.

### DIFF
--- a/lib/APPTurtle/src/CustomRosTransport.cpp
+++ b/lib/APPTurtle/src/CustomRosTransport.cpp
@@ -38,6 +38,7 @@
 #include <WiFiUdp.h>
 #include <SimpleTimer.hpp>
 #include <Logging.h>
+#include <Board.h>
 
 /******************************************************************************
  * Compiler Switches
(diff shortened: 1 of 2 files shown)


lib/APPTurtle/src/CustomRosTransport.cpp
```diff
@@ -189,6 +190,11 @@ size_t custom_transport_read(uxrCustomTransport* transport, uint8_t* buffer, siz
             {
                 break;
             }
+
+/* ROS executor requires active time. Added manual simulation time step for now. */
```
nhjschulz, line 194: move into #if

nhjschulz (commented):
Iboard.h missing ?

lib/APPTurtle/src/CustomRosTransport.cpp
hoeftjch: done

nhjschulz (approved):

PR #173, opened 2024-10-15
ROS2 Tutorial improvements

BlueAndi:
Improvements after hands-on setup.


lib/APPTurtle/src/MicroRosClient.cpp
```diff
@@ -224,6 +227,8 @@ bool MicroRosClient::configureClient()
     {
         isSuccessful = true;
 
+        LOG_INFO("Conneced with micro-ROS agent.");
```
gabryelreyes, line 230: Typo in connected

gabryelreyes (approved):

lib/APPTurtle/src/MicroRosClient.cpp
BlueAndi: Fixed

doc/ROS2/setup/Webots.md
```diff
@@ -7,6 +7,14 @@ Follow the official installation guide for Webots on Linux with APT on the
 > **_NOTE:_**
 Some of the steps result in larger package downloads.
 
+Set environment variables for Webots home directory, the Webots controller library directory and add the Webots controller library to linker search path.
```
nhjschulz, line 10: Add a sentence that setting them in ".bashrc" is recommended to have it in any shell automatically. It reads as it is a "must" right now.  

Some users are picky regarding changes in this file and prefer other means to set this up.

doc/ROS2/setup/microROS_Native.md
```diff
@@ -28,7 +29,6 @@ colcon build
 source install/local_setup.bash
 ```
 
-
 ## Create static micro-ROS library
```
nhjschulz, line 32: suggest to add "for DroidControlShip" to the heading 

This makes it clearer that we are now  using stuff from this Repo, not microROS.

Also a reasoning sentence whould be helpful here, like :

*The static library is needed as DroidControlShip is build outside of the ROS2 environment with PlatformIO and.*

doc/ROS2/setup/microROS_Native.md
```diff
@@ -39,7 +39,9 @@ ros2 run micro_ros_setup create_firmware_ws.sh generate_lib
 ```
 
 Copy the files [my_custom_toolchain.cmake](./native_build/my_custom_toolchain.cmake) and [my_custom_colcon.meta](./native_build/my_custom_colcon.meta) in the microros_ws
-
+```bash
+cp ~/DroidControlShip/doc/ROS2/setup/native_build/my_custom_* ~/microros_ws
```
nhjschulz, line 43: This "\~" only works if DCS in directly under $HOME. Some user may not know what "\~" does and fail.
How about 
Example (replace \<DCS-DIR\> with the DroidControlShip workspace location):

`cp <DCS-DIR>/doc/...`

lib/APPTurtle/src/MicroRosClient.cpp
```diff
@@ -198,6 +198,9 @@ bool MicroRosClient::configureClient()
     rcl_init_options_t init_options = rcl_get_zero_initialized_init_options();
     m_allocator                     = rcl_get_default_allocator();
 
+    LOG_INFO("Connecting to micro-ROS agent %s:%u ...", m_agentConfiguration.address.toString().c_str(),
```
nhjschulz, line 201: The pattern `toString().c_str()` ist risky.  The "toString()" method returns a temporary string that nobody uses after calling c_str(). It might get destructed right after the c_str() call, making the pointer dangling.  

Recommend to use safe pattern:

         String ipAddrStr(m_agentConfiguration.address.toString());
         LOG_INFO("Connecting to micro-ROS agent %s:%u ...",
                  ipAddrStr.c_str(),
                  m_agentConfiguration.port);

nhjschulz (changes requested):

doc/ROS2/setup/microROS_Native.md
BlueAndi: Following the tutorial it is located in the home directory. Otherwise in all other places this must be replaced too.

If someone has Linux knowledge, he knows what to do, the others are happy for every clear and simple "just do it" line.

nhjschulz (approved):

Issue #176, opened 2024-10-19
Implement Micro-ROS agent retry mechanism


BlueAndi:
The Micro-ROS agent connection establishment takes currently only place once. If the Micro-ROS agent is e.g. restarted, DCS will die and needs a reset.

Issue #175, opened 2024-10-19
Remove all infinite loops


BlueAndi:
Using a infinite loop in case of fatal errors makes the system non-responsive in the native environment. Therefore the infinite loops shall be replaced appropriate.

PR #177, opened 2024-10-21
Micro-ROS client retry mechanism

BlueAndi:
#175
#176

### DIFF
--- a/lib/APPTurtle/src/CustomRosTransport.cpp
+++ b/lib/APPTurtle/src/CustomRosTransport.cpp
@@ -71,8 +71,8 @@ bool CustomRosTransport::open(uxrCustomTransport* transport)
     }
     else
     {
-        const CustomRosTransport* tthis  = reinterpret_cast<CustomRosTransport*>(transport->args);
-        const int                 UDP_OK = 1;
+        CustomRosTransport* tthis  = reinterpret_cast<CustomRosTransport*>(transport->args);
+        const int           UDP_OK = 1;
 
         if (UDP_OK != tthis->m_udpClient.begin(tthis->m_port))
         {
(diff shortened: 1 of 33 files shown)


lib/APPTurtle/src/MicroRosClient.cpp
```diff
@@ -178,12 +124,39 @@ bool MicroRosClient::registerSubscriber(BaseSubscriber* subscriber)
...
+    case STATE_CONNECTED:
+        connectedState();
+        break;
+
+    case STATE_DISCONNECTED:
+        disconnectedState();
+        break;
+
+    default:
+        /* Should never happend. */
```
gabryelreyes, line 155: typo "happen"

gabryelreyes (approved):

gabryelreyes (approved):

lib/APPTurtle/src/CustomRosTransport.cpp
```diff
@@ -77,10 +71,10 @@ bool CustomRosTransport::open(uxrCustomTransport* transport)
     }
     else
     {
-        const micro_ros_agent_locator* locator = reinterpret_cast<micro_ros_agent_locator*>(transport->args);
-        const int                      UDP_OK  = 1;
+        CustomRosTransport* tthis  = reinterpret_cast<CustomRosTransport*>(transport->args);
```
nhjschulz, line 74: I never saw this  tthis name convention before and first thought it is a typo.
Can't we be more explicit and call it "transport" or "thisTransport" to be closer to naming CS.

nhjschulz (commented):
optional comment on variable name

lib/APPTurtle/src/CustomRosTransport.cpp
BlueAndi: We can fix it with next feature integration, this PR is already merged.

PR #180, opened 2024-10-22
MircoRosTransport: Add C++ methods

nhjschulz:
Add c++ transport handler and  make the C-functions entry points only wrappers for the new c++ ones.

This removes the need for a "handcrafted" this pointer inside the C-methods for accessing Class elements.

### DIFF
--- a/lib/APPTurtle/src/CustomRosTransport.cpp
+++ b/lib/APPTurtle/src/CustomRosTransport.cpp
@@ -53,6 +53,18 @@
  * Prototypes
  *****************************************************************************/
 
+/**
+ * Unwrap pointer to CustomRosTransport from uxrCustomTransport structure.
+ * 
+ * This is used by the static C-Languge entry points to forward the request
+ * to the matching C++ member function.
+ *
+ * @param[in] transport The custom transport data structure pointer.
+ *
+ * @return The this pointer to transport owning CustomRosTransport class.
+ */
+static CustomRosTransport* toThis(const uxrCustomTransport* transport);
+
 /******************************************************************************
  * Local Variables
  *****************************************************************************/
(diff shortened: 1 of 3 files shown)


lib/APPTurtle/src/CustomRosTransport.cpp
```diff
@@ -63,81 +75,111 @@
...
+{
+    CustomRosTransport * tp = toThis(transport);
+    return (nullptr != tp) ? tp->read(buffer, size, timeout, errorCode) : 0U;
+}
+
+/******************************************************************************
+ * Local Functions
+ *****************************************************************************/
+
+static CustomRosTransport* toThis(const uxrCustomTransport* transport)
```
BlueAndi, line 105: Add "inline" to the local function.

lib/APPTurtle/src/CustomRosTransport.cpp
```diff
@@ -63,81 +75,111 @@
...
 
+size_t CustomRosTransport::read(uxrCustomTransport* transport, uint8_t* buffer, size_t size, int timeout,
+                                uint8_t* errorCode)
+{
+    CustomRosTransport * tp = toThis(transport);
+    return (nullptr != tp) ? tp->read(buffer, size, timeout, errorCode) : 0U;
+}
+
+/******************************************************************************
+ * Local Functions
```
BlueAndi, line 102: Local functions are ususally behind the method sections.

BlueAndi (changes requested):

lib/APPTurtle/src/CustomRosTransport.cpp
nhjschulz: done

BlueAndi: push is missing.

lib/APPTurtle/src/CustomRosTransport.cpp
nhjschulz: done

BlueAndi (approved):

PR #203, opened 2025-11-04
State Diagram

haecto:
Pretty unsure about the State diagram, as there will be no changes in use of SensorFusion

Anyway did it for documentation, please review


doc/SensorFusion/StateDiagram.puml
```diff
@@ -0,0 +1,32 @@
+@startuml StateDiagram
```
jkerpe: Is this really relevant for your thesis?

PR #210, opened 2025-12-27
Support Hardware V2

gabryelreyes:
USB Host driver is not implemented yet. Will be done in a separate PR. See #217


BlueAndi (changes requested):
There is much more code that could be in a common HAL to keep the specific HAL as small as possible.

BlueAndi (approved):

PR #221, opened 2026-02-23
Improve ROS2 doc

stefanvogel:
What remains after rebasing my branch onto feature/ros2Wsl.

Sorry for the branch name, which results from the includes Fix from last week, which you have found meanwhile on feature/ros2Wsl as well.

### DIFF
--- a/lib/APPTurtle/src/App.cpp
+++ b/lib/APPTurtle/src/App.cpp
@@ -88,7 +88,7 @@ void App::setup()
     /* Register serial log sink and select it per default. */
     if (true == Logging::getInstance().registerSink(&gLogSinkSerial))
     {
-        (void)Logging::getInstance().selectSink("Serial");
+        (void)Logging::getInstance().selectSink(gLogSinkSerial.getName());
 
         /* Set severity of logging system. */
         Logging::getInstance().setLogLevel(CONFIG_LOG_SEVERITY);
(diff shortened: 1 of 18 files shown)


lib/APPTurtle/src/App.cpp
```diff
@@ -296,7 +296,7 @@ void App::handleTurtle()
         RobotSpeed    payload;
         const int32_t MILLI_CONVERSION_FACTOR = 1000;
         int32_t       linearSpeed = m_turtleSpeedSetpoint.linear.x * MILLI_CONVERSION_FACTOR; /* Linear speed in mm/s */
-        int32_t       angularSpeed = m_turtleSpeedSetpoint.angular.z * MILLI_CONVERSION_FACTOR; /* Angular speed in mrad/s */
+        int32_t angularSpeed = m_turtleSpeedSetpoint.angular.z * MILLI_CONVERSION_FACTOR; /* Angular speed in mrad/s */
```
BlueAndi: Why moving back?

BlueAndi (changes requested):

lib/APPTurtle/src/App.cpp
stefanvogel: clang-format automatically does

BlueAndi (approved):

PR #223, opened 2026-02-25
Improve github workflow

stefanvogel:


.github/workflows/main.yml
```diff
@@ -1,43 +1,34 @@
...
   intro:
-    # The type of runner that the job will run on.
     runs-on: ubuntu-24.04
-    # Steps represent a sequence of tasks that will be executed as part of the job.
+
     steps:
     - run: echo "The job was automatically triggered by a ${{ github.event_name }} event."
     - run: echo "The name of the branch is ${{ github.ref }} and the repository is ${{ github.repository }}."
 
-    - name: Checkout repository
```
gabryelreyes, line 24: Doesnt running the setup in the intro speed up the running of other jobs, as the cache is then updated for all dependent jobs?

gabryelreyes (approved):

.github/workflows/main.yml
stefanvogel: No — running setup steps in the intro job would not speed up the other jobs, and it would not share an updated cache with them. Nothing installed in intro persists to other jobs.

Issue #186, opened 2024-10-31
Coding guidelines are  missing


nhjschulz:
Issue: 
The project is lacking coding guidelines, only formatting rules exist through clang-format.
"Pathfinder rule" is applied, but works well only for style formatting, not regarding langugage feature usage.

Impact:
* Lack of rules leads to review feedback biased to personal preferences.
* No reasoning why a construct should be used or not
* Inconsistencies when to add/remove optional C++ syntax elements that are implicit if missing .

This impairs collaboration when the number of comitters grows.

Proposed solution:
* Create/Link to coding guideline rules and improve them over time based on code review findings.

PR #228, opened 2026-03-06
Coding guideline

BlueAndi:
#186

### DIFF
--- a/doc/CodingGuideline.md
+++ b/doc/CodingGuideline.md
@@ -1,8 +1,23 @@
-# Coding Guideline
+# Coding Guideline <!-- omit in toc -->
 
 Derived from existing source code in this repository.
 
----
+## Table of Contents <!-- omit in toc -->
+
+- [File Structure](#file-structure)
+- [Naming](#naming)
+- [Headers](#headers)
+- [Doxygen](#doxygen)
+- [Class Layout](#class-layout)
+- [Patterns](#patterns)
+  - [Singleton](#singleton)
+  - [State Machine](#state-machine)
+- [Formatting](#formatting)
+- [Conditionals](#conditionals)
+  - [Switch](#switch)
+- [Parameters](#parameters)
+- [Logging](#logging)
+- [Tests](#tests)
 
 ## File Structure
 
@@ -48,8 +63,6 @@ Each section uses a 80-character comment banner:
 Close every header file with:
 
 ```cpp
-#endif /* FILENAME_H */
-
 /** @} */
 ```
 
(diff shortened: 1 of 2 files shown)


doc/CodingGuideline.md
```diff
@@ -0,0 +1,257 @@
...
+
+Each section uses a 80-character comment banner:
+
+```cpp
+/******************************************************************************
+ * Includes
+ *****************************************************************************/
+```
+
+Close every header file with:
```
nhjschulz, line 63: There seems to be the header start missing.  it jump from cpp source to  ending header. file. The header start with guard should be in between.

doc/CodingGuideline.md
```diff
@@ -0,0 +1,257 @@
...
+
+| Element | Convention | Example |
+|---|---|---|
+| Class | PascalCase | `StateMachine`, `StartupState` |
+| Interface / abstract base | Prefix `I` | `IState`, `IBoard` |
+| Method | camelCase | `setup()`, `getInstance()` |
+| Private member variable | `m_` prefix + camelCase | `m_isActive`, `m_duration` |
+| Constant / `static const` member | `UPPER_SNAKE_CASE` | `MIN_BATTERY_LEVEL`, `SEND_WAYPOINT_TIMER_INTERVAL` |
+| File-scope static variable | `g` prefix + camelCase | `gLogSinkSerial` |
+| Macro parameter | double-underscore prefix | `SIMPLE_TIMER_SECONDS(__timeInS)` |
```
nhjschulz: Is this already used ?  Dunder prefix is reserved to compiler/runtime.

doc/CodingGuideline.md
```diff
@@ -0,0 +1,257 @@
...
+| Enum value | `UPPER_SNAKE_CASE` | `CMD_GET_MAX_SPEED`, `CMD_NONE` |
+
+Use `uint8_t`, `uint16_t`, `uint32_t` etc. for sized integer types. Append `U` to unsigned integer literals (`50U`, `1000U`).
```
nhjschulz, line 84: Wouldn't it simpler to refer to stdint.h base type usage.

doc/CodingGuideline.md
```diff
@@ -0,0 +1,257 @@
...
+
+---
+
+## Headers
```
nhjschulz, line 88: That should better move up to appear right after (or before) cpp skeleton.

doc/CodingGuideline.md
```diff
@@ -0,0 +1,257 @@
...
+// ...
+#endif /* STATE_MACHINE_H */
+```
+
+Template/header-only implementations use `.hpp`, all others use `.h` / `.cpp`.
+
+Include order (within a file):
+1. Own header (`"App.h"`)
+2. Project headers (`<Board.h>`, `<Logging.h>`)
+3. Third-party / Arduino headers (`<ArduinoJson.h>`)
```
nhjschulz, line 104: std headers ?

doc/CodingGuideline.md
```diff
@@ -0,0 +1,257 @@
...
+```cpp
+#ifndef STATE_MACHINE_H
+#define STATE_MACHINE_H
+// ...
+#endif /* STATE_MACHINE_H */
+```
+
+Template/header-only implementations use `.hpp`, all others use `.h` / `.cpp`.
+
+Include order (within a file):
```
nhjschulz, line 101: It may make sense to distinguish header and cpp usages here.
i.e a header should only include the minimum to allow compilation, not everything its module needs in implementation.

doc/CodingGuideline.md
```diff
@@ -0,0 +1,257 @@
...
+1. Constructors / destructor
+2. Static factory / `getInstance()`
+3. Public methods
+4. Static constants
+5. Member variables
+6. Deleted copy constructor / assignment operator
+
+---
+
+## Patterns
```
nhjschulz, line 147: for me such patterns are not desirable in a style guide. A singleton pattern has a lot more dependencies and quirks than style. 
I also don't get the fous for "final" for concrete states. Better would be a rule when to apply final (and when not).

doc/CodingGuideline.md
```diff
@@ -0,0 +1,257 @@
...
+default:
+    break;
+}
+```
+
+---
+
+## Parameters
+
+Pass non-trivial input arguments as `const` reference; output arguments as non-const reference:
```
nhjschulz, line 247: Any rule regarding pointer passing ?  References are not working if a i nput/output is optional, i.e. may be a nullptr.

nhjschulz (changes requested):
Some comments, mostly subjective as this is style guide related. Feel free to reject if you disagree.

PR #230, opened 2026-04-08
feat(RemoteControl): add IMU data

stefanvogel:


lib/APPRemoteControl/src/App.cpp
nhjschulz, line 472: Is there  a need for backward compatibility based on playload size being less than the full data set ?

stefanvogel: Backward compatibility is not feasible here because `timestamp` was prepended rather than appended to the `VehicleData` struct in RadonUlzer, which shifts all existing field offsets.

Hence both sides must run matching firmware.

See https://github.com/BlueAndi/RadonUlzer/commit/4cd8cc42f0ebb58653dad0eae9041dec0837b2e6#diff-cdca5922850e1cb2ee03ab5e93848bc61d76bfd0abee516d9610601817c085d8

BlueAndi: Appending would cause a problem anyway. :-) DCS must follow RU here in this case.

nhjschulz (approved):

